PR #93, opened 2022-08-05
Custom DRF frontend

dandavies99:
This PR adds work by @pxjacomex to add a custom `rest_framework/api.html` so that there is a more useful browsable frontend than the one that ships with DRF, including a menu bar etc. This can be used as a temporary solution to a frontend before the React frontend. 

@drklrd - Before Pablo makes any more additions to this, can you please review the changes and make sure this will not be a problem from your end? I assume it will be fine because only the html template is being modified, the API behaviour itself is the same. 

@pxjacomex - my only suggestion: Is it possible to keep the dropdown button so we can view the dictionary (api) version of the GET response? E.g. in the image below, perhaps we could have 3 option, [JSON, API, TABLE.] ?

image


dandavies99:
Hi @pxjacomex, thanks for the changes. I'm not sure if we need a separate HTML response though - I think your table response is great. 

It would be better to retain the existing api view as well, rather than replace it entirely with this new table view. I think the best way is to [create a custom renderer](https://www.django-rest-framework.org/api-guide/renderers/#custom-renderers) for the table response, then add that to the `DEFAULT_RENDERER_CLASSES`. This will use the custom api.html that you modified (can rename to table.html), and the BrowsableAPIRenderer can use the default api.html.

So overall we can have:

- "rest_framework.renderers.JSONRenderer" - No template needed 
   "rest_framework.renderers.BrowsableAPIRenderer" - default api.html template 
   "djangomain.renderers.TablerAPIRenderer" - custom table.html template 

Does that make sense?

pxjacomex:
> Hi @pxjacomex, thanks for the changes. I'm not sure if we need a separate HTML response though - I think your table response is great.
> 
> It would be better to retain the existing api view as well, rather than replace it entirely with this new table view. I think the best way is to [create a custom renderer](https://www.django-rest-framework.org/api-guide/renderers/#custom-renderers) for the table response, then add that to the `DEFAULT_RENDERER_CLASSES`. This will use the custom api.html that you modified (can rename to table.html), and the BrowsableAPIRenderer can use the default api.html.
> 
> So overall we can have:
> 
> * "rest_framework.renderers.JSONRenderer" - No template needed
>   "rest_framework.renderers.BrowsableAPIRenderer" - default api.html template
>   "djangomain.renderers.TableAPIRenderer" - custom table.html template
> 
> Does that make sense?

@dandavies99 Ok. I gonna make it using a custom renderer. Thanks for the explanation.

pxjacomex:
> Thanks @pxjacomex - these updates look good. Do you think this is ready to merge?

@dandavies99 You're welcome. I would like to do some checks before merging it. I think that will be ready next week.

dandavies99:
Hi @pxjacomex - the behaviour after updating/changing/deleting works well for me. 👍

PR #128, opened 2024-01-16
Add MeasurementBase model

### LINKED ISSUE #114: Add a base measurement model
This should be a rework (in a separate class, for now) of the existing [`BaseMeasurement` model and the `create_meas_model` function](https://github.com/ImperialCollegeLondon/paricia/blob/60a30b688977a2332d96b01828fd310b4b9e75d5/measurement/models.py#L116). 

This class should contain the basic information required by all measurement related models, namely `Measurement` and `Report`. As such it will contain all the relevant fields that are not specific for any of those. See relevant issues for those models (#112 and #XXX). 

- As this model combine all the raw measurement variables (28) it might have slightly different fields, these fields might need to be optional and left blank, so only those variables that really need them have to populate them.
- The magnitude should always be "value" and decide what to do with it in the reports based on the `Variable.is_cumulative` flag (maybe we could add a label to `Variable` to indicate this: `value`, `average` or `sum`).
- It should be an abstract class, so no table is created for it.

dalonsoa:
Close #114 

Adds the base model that will be used by all measurement-related models. It will replace, ultimately, the existing `BaseMeasurement` class and `create_meas_model` function. 

As it is an abstract model that do not add an entry to the DB and , there's nothing to test (at least [not without a lot of work](https://stackoverflow.com/questions/4281670/django-best-way-to-unit-test-an-abstract-model)). Tests will be added down the line, after benchmarking has proved that this is performant enough.

### DIFF
--- a/measurement/models.py
+++ b/measurement/models.py
@@ -19,9 +19,54 @@ from django.urls import reverse
 from timescale.db.models.models import TimescaleModel
 
 from station.models import Station
+from variable.models import Variable
 
 MEASUREMENTS: List[str] = []
 """Available measurement variables."""
+
+
+class MeasurementBase(TimescaleModel):
+    """Base class for all the measurement related entries.
+
+    It contains the barebone attributes that any measurement entry will likely need,
+    although this is enforced only for station, variable and value. Maximum and minimum
+    are very likely to be present in most cases, but might not be there in some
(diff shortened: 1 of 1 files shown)


tsmbland (approved):
Looks good!

PR #130, opened 2024-01-16
Adds the new report model

### LINKED ISSUE #115: Add a `Report` model
This model holds the information related to hourly, daily and monthly reports. It inherits from `MeasurementBase` (#114) and has the following additional fields:

- `report_type`: Text field. Can be one of `hourly`, `daily` or `monthly`
- `used_for_daily`: Boolean. Defaults to False. Can only become True if `report_type` is `hourly` as only `hourly` data can be used to generate `daily` reports.
- `used_for_monthly`: Boolean. Defaults to False. Can only become True if `report_type` is `daily` as only `daily` data can be used to generate `monthly` reports.
- Indexing is going to be critical for performance. We might need to tweak this during benchmarking, but a reasonable indexing could be `(station, variable, report_type, time)`. It depends on what the most common queries are.

The relationship between the different flags to ensure they are correct (eg. `used_for_daily` is true while `report_type` is monthly) will need to be assessed and enforced in the `clean` method of the model.

dalonsoa:
Close #115 

Ads the new Report model, inheriting from `MeasurementBase` (see #128 ), including the report type, flags to indicate when it has been used and validation. 

For the tests, the package `model_bakery` has been added, and the CI the workflow file has been updated accordingly. I've also added this to the Dockerfile as we are in the development phase. Ideally, we will want to get rid of it in production, so the image do not include unnecessary packages, but it should be ok for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip wheel
         pip install -r requirements.txt
+        pip install -r requirements-dev.txt
     - name: Run migrations
       run: python manage.py migrate
     - name: Run tests
--- a/measurement/models.py
+++ b/measurement/models.py
@@ -63,10 +63,49 @@ class MeasurementBase(TimescaleModel):
         default_permissions = ()
         abstract = True
         indexes = [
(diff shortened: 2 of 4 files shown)


tsmbland (approved):
Looks good

PR #131, opened 2024-01-17
Adds the new Measurement model

### LINKED ISSUE #112: Write a measurement model
This will hold the raw and validated data and derives from the new base class created in #114 . It has the following specific fields:

- `is_validated`: Boolean, indicates if the data has been validated.
- `used_for_hourly`: Boolean, indicates if the data has been used for hourly reports. Default to False. Only can be True if `is_validated` is True as only validated data can be used for `hourly` reports.
- `raw`: Is empty if the data is not validated or if it is validated and a new value has been used to override the existing one. In this case, it holds the original value. It might be necessary to include several `raw` values, eg. `raw_maximum`, `raw_minimun`, etc. if those can also be manually changed. 
- `is_active`: Boolean, indicate if the data is usable for further processing. Defaults to True. Only validated data can set this flag to False.  
- Indexing is going to be critical for performance. We might need to tweak this during benchmarking, but a reasonable indexing could be `(station, variable, time, is_validated)`. It depends on what the most common queries are.

The relationship between the different flags to ensure they are correct (eg. `it_active` is false while `is_validated` is also false) will need to be assessed and enforced in the `clean` method of the model.

dalonsoa:
Close #112 

To be reviewed and merged after #130 

This class holds the value of a given variable and station at a specific time, as well as auxiliary information such as maximum and minimum values, depth and direction, for vector quantities. All of these hava a `raw` version where a backup of the original data is kept, should this change at any point.

Flags to monitor its validation status, if the data is active (and therefore can be  used for reporting) and if it has actually been used for that is also included.


tsmbland (approved):
Looks good

PR #133, opened 2024-01-25
Add completeness and time validation to Report

### LINKED ISSUE #132: Add `completeness` field to Report and time rounding to appropriate value
The Report model is missing a completeness field that indicates the fraction of the period that had data. Eg. if I only have 18h of data out of 24h, then this filed will indicate 0.75. That's calculated somewhere else, but the field needs to exist to store that data. 

Also, for consistency, dates and times of the reports should be consistent. This means hourly data should have minutes.seconds equal to 00.0, daily data should be reported for hour 00:00.0 and monthly data for day 1 (and 00:00.0). This can be enforced in the `clean` method.

dalonsoa:
Close #132

### DIFF
--- a/measurement/models.py
+++ b/measurement/models.py
@@ -74,6 +74,7 @@ class Report(MeasurementBase):
     used_for_monthly = models.BooleanField(
         verbose_name="Used for monthly?", default=False
     )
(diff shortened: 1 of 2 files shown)


tsmbland (approved):
Looks good

Issue #121, opened 2023-12-21
Replace JS code for validation (umbrella issue)


dalonsoa:
Like with the reports (#120) the validation screen is currently created and displayed with a mixture of:

- [Forms](https://github.com/ImperialCollegeLondon/paricia/blob/60a30b688977a2332d96b01828fd310b4b9e75d5/validated/forms.py#L8)
- Views:
  - [DailyValidation](https://github.com/ImperialCollegeLondon/paricia/blob/60a30b688977a2332d96b01828fd310b4b9e75d5/validated/views.py#L288)
  - [DetailList](https://github.com/ImperialCollegeLondon/paricia/blob/60a30b688977a2332d96b01828fd310b4b9e75d5/validated/views.py#L318)
- [HTML templates](https://github.com/ImperialCollegeLondon/paricia/tree/60a30b688977a2332d96b01828fd310b4b9e75d5/validated/templates)
- [JS code to create the plots](https://github.com/ImperialCollegeLondon/paricia/blob/develop/validated/static/validated/daily_validation.js)

With diverse logic here and there to deal with the data and the user interaction. While a variety of files is normally required in Django, this is a mess.

This task will have as main goal removing JS code altogether and replace it with appropriate equivalents in terms of content and user interaction using **Dash/Plotly**. These will need to be hooked in the correct places of the views and/or forms, and the HTML template modified accordingly.

This might be a substantial task and should be broken down into smaller issues as needed, keeping this as umbrella.

![Screenshot from 2024-01-09 07-51-59](https://github.com/ImperialCollegeLondon/paricia/assets/6095790/84b5515e-2213-4c4b-93b7-f7cfb090e7cd)
![Screenshot from 2024-01-09 07-51-51](https://github.com/ImperialCollegeLondon/paricia/assets/6095790/cf1e24be-91ff-4cac-880a-4f6423b9050e)
![Screenshot from 2024-01-09 07-52-04](https://github.com/ImperialCollegeLondon/paricia/assets/6095790/1eebe119-aa24-4fc8-86b6-afc09630f4b1)

tsmbland:
A few notes and ideas, but would definitely want to discuss first before proceeding.

-----

OVERALL PLAN

1. Modify project to be compatible with dash apps using `django_plotly_dash` (Test this approach first using a minimal 'hello world' dash app). Consult with Callum.
	- https://django-plotly-dash.readthedocs.io/en/latest/
[...]

PR #143, opened 2024-01-30
Create tables for Validation app

### LINKED ISSUE #138: Convert tables from javascript to python (plotly/dash)
For #121

tsmbland:
Continuing with #121. This PR is just about setting the layout of the tables, showing the right data, and conditional colouring of cells. I still need to write callbacks and buttons to make this functional, and decide on the overall layout of the app

Close #138


validated/dash_apps/finished_apps/daily_validation.py
```diff
@@ -30,10 +32,67 @@
...
+        mode="lines",
+        name="Validated",
+        line=dict(color="#00CC96"),
+    )
+)
 
-# Create layout
-app.layout = html.Div([dcc.Graph(figure=plot)])
+# Layout
+app.layout = html.Div(
```
dalonsoa, line 80: I think this results in an odd number of vertical scroll bars.

- I've the external one - the normal one of the browser.
- The internal one to scroll a single table.
- An intermediate one that apparently scrolls through the two tables and the plot. 

I don't think I see the point of this intermediate scrollbar and, moreover, it makes the navigation rather complicated. Is there a way of removing it and leave just two (the browser one and the table specific one)?.
Screenshot from 2024-02-05 17-27-55

validated/tables.py
```diff
@@ -0,0 +1,287 @@
...
+    table = AgGrid(
+        id="table_daily",
+        rowData=data["data"],
+        columnDefs=create_columns_daily(value_columns=data["value_columns"]),
+        columnSize="sizeToFit",
+        defaultColDef={
+            "resizable": True,
+            "sortable": True,
+            "checkboxSelection": {
+                "function": "params.column == params.columnApi.getAllDisplayedColumns()[0]"
```
dalonsoa, line 22: Where is this `params` object coming from? Or, in other words, how this `checkboxSelection` business work?

validated/dash_apps/finished_apps/daily_validation.py
```diff
@@ -1,15 +1,17 @@
...
+from validated.functions import daily_validation, detail_list
+from validated.tables import create_daily_table, create_detail_table
 from variable.models import Variable
 
-# Create a Dash app
+DEFAULT_FONT = "Open Sans, Raleway, Dosis, Ubuntu, sans-serif"
+
 app = DjangoDash("DailyValidation")
 
 # Filters (in final app this will get data from forms)
```
dalonsoa, line 17: I'm not logged in and all this DB querying works fine. Nothing to worry about at this time, as this part is under development, but at some point this will need to go thorugh the authentication system - specially the validation view, not so much @CWestICL 's reporting view.

dalonsoa (changes requested):
This looks good, although `dash` isn't super easy to read. My only somewhat major comment at this time is the scrollbars one: there are too many! Is there anything we can do about it?

Also, I thought you were planning to put the tables in tabs. Whatever you choose is ok, just checking if you changed your mind.

validated/dash_apps/finished_apps/daily_validation.py
tsmbland: I'll add this as a separate issue

validated/dash_apps/finished_apps/daily_validation.py
tsmbland: This seems to have fixed itself in a later PR

PR #144, opened 2024-02-01
Add synthetic data creation scenarios

dalonsoa:
Adds the scripts to create synthetic data in two scenarios, ~ 24M entries spread over 23 years and the same entries compressed in just 1 year. 

A Wiki summarising the findings is included here: https://github.com/ImperialCollegeLondon/paricia/wiki/Performance-benchmarking

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,120 +7,121 @@ db.sqlite3
 media
 */db/*
 
-# Migrations 
+# Migrations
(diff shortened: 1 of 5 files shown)


tsmbland (approved):
Looks good. Gave it a quick try and everything works (although I didn't run it to completion)

PR #145, opened 2024-02-01
Add callbacks for adding new data and saving to validated

### LINKED ISSUE #140: Callbacks for saving validated data
For #121 

To do after #138

### LINKED ISSUE #148: Ability to add new rows to the detail table

tsmbland:
To review after #143 

This adds some capabilities to the daily validation app using callbacks.

Current capabilities:
- Manually editing values in the detail table
- Saving selected entries to Validated
- Resetting Validated (removes all entries in the table from Validated and resets edited values)
- Adding new rows to the detail table

Notes:
- I'm doing this by calling Python functions from functions.py directly rather than using the API. Not sure what's more appropriate
- I had to move the table building outside of a function and into the main dash app file to make them work with callbacks

Close #140 
Close #148

### DIFF
--- a/validated/dash_apps/finished_apps/daily_validation.py
+++ b/validated/dash_apps/finished_apps/daily_validation.py
@@ -183,6 +183,8 @@ app.layout = html.Div(
         Output("table_detail", "rowData"),
         Output("table_detail", "rowTransaction"),
         Output("table_detail", "scrollTo"),
+        Output("table_daily", "selectedRows"),
+        Output("table_detail", "selectedRows"),
     ],
     [
         Input("daily-save-button", "n_clicks"),
@@ -209,7 +211,17 @@ def buttons_callback(
     in_daily_row_data: list[dict],
     in_detail_selected_rows: list[dict],
     in_detail_row_data: list[dict],
-) -> tuple[str, str, go.Figure, list[dict], list[dict], list[dict], list[dict]]:
+) -> tuple[
+    str,
+    str,
+    go.Figure,
+    list[dict],
+    list[dict],
+    dict,
+    dict,
+    list[dict],
+    list[dict],
+]:
     """Callback for buttons adding and resetting Validated data
 
     Args:
@@ -223,7 +235,7 @@ def buttons_callback(
         in_detail_row_data (list[dict]): Full row data for table_detail
 
     Returns:
-        tuple[str, str, go.Figure, list[dict], list[dict], dict, dict]:
+        tuple[str, str, go.Figure, list[dict], list[dict], dict, dict, list[dict], list[dict]]:
             Callback outputs
     """
     global DATA_DAILY, DATA_DETAIL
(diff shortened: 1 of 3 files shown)


validated/dash_apps/finished_apps/daily_validation.py
```diff
@@ -84,15 +145,205 @@
...
+    # Button: Detail reset
+    elif button_id == "detail-reset-button":
+        reset_detail_validated(
+            data_list=in_detail_row_data,
+            variable=variable,
+            station=station,
+        )
+        out_detail_status = "Validation reset"
+        plot_refresh_required = True
+        detail_refresh_required = True
```
dalonsoa, line 297: When clicking this, part of the validated plot disapears, but nothing to do with the selected/unselected range.

validated/tables.py
```diff
@@ -1,66 +1,7 @@
-from dash_ag_grid import AgGrid
-
-
-def create_daily_table(data: dict) -> AgGrid:
```
dalonsoa, line 4: Why these two need to be removed from there? Is it a dash thing?

dalonsoa (changes requested):
I'm not sure if I'm doing something wrong, but it seems to me the buttons are not doing what they are expected to do, but I'm not sure if because of the logic of the buttons or some other underlaying issue. I'm recorded a video (unfortunately, the mouse pointer is not shown, but hopefully you can follow).

[Screencast from 06-02-24 05:59:41.webm](https://github.com/ImperialCollegeLondon/paricia/assets/6095790/d3aa3417-dd18-4bc7-b25e-a6ae777ec455)

Also, when I selected a row in the daily table, I do not see a change in the detail table. 

I'm glad the extra scrolbar I mentioned in the other PR is gone now. 

Is there a way of indicating what day we are seeing in the detail table, addding it to the title or something?

validated/dash_apps/finished_apps/daily_validation.py
```diff
@@ -84,15 +145,205 @@
...
+
+    Returns:
+        tuple[str, str, go.Figure, list[dict], list[dict], dict, dict]:
+            Callback outputs
+    """
+    global DATA_DAILY, DATA_DETAIL
+
+    ctx = dash.callback_context
+    if not ctx.triggered:
+        return (dash.no_update,) * 7
```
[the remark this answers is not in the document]
tsmbland: This was just because callbacks are usually triggered when the app first loads so this would prevent anything happening in that case. However, I also have `prevent_initial_call=True` so this is redundant and I will remove it.

validated/dash_apps/finished_apps/daily_validation.py
tsmbland: This is by design, as the reset buttons are intended to remove the entire table from validated, rather than just the selected rows (so if you mess up and want to start again). I could add a separate "Remove from Validated" button to remove just the selected rows if helpful?

validated/tables.py
tsmbland: Yeah, a problem with the callback not recognizing the object. It's ok to build objects with functions, but for some reason those functions need to be in the same file as the callback or it has problems. Ran into similar issues with Gridlington.

PR #146, opened 2024-02-02
Data Report dash app

CWestICL:
An attempt to replace the graph section in the current `data_report.html` template with a newly created dash app imported from `data_report.py`.

My intention is to fit the new dash app into the current `DataReport(FormView)` so that the form data can be passed into the app as `initial_arguments` and used to populate the graph. Currently the `DataReportDash` app is a copy of the `DailyValidation` app.

I have tried using `from .dash_apps.finished_apps import data_report` in the `DataReport(FormView)` as was done for the `DailyValidationDev(View)` but I still get an error that the `DataReportDash` app cannot be found.

### DIFF
--- a/validated/views.py
+++ b/validated/views.py
@@ -400,6 +400,8 @@ class DataReport(FormView):
     success_url = "/validated/data_report/"
 
     def post(self, request, *args, **kwargs):
+        from .dash_apps.finished_apps import data_report
+        
         form = DataReportForm(self.request.POST or None)
         if form.is_valid():
             if functions.is_ajax(self.request):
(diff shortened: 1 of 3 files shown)


validated/views.py
```diff
@@ -400,6 +400,8 @@ class DataReport(FormView):
     success_url = "/validated/data_report/"
 
     def post(self, request, *args, **kwargs):
+        from .dash_apps.finished_apps import data_report
```
dalonsoa, line 403: I'm not entirely sure, but this is a `post` request, so it not executed just when requesting the page, but when selecting something in the existing form (the station, variable, etc.) and asking for the report. As such, when you just load `/validated/data_report/` for the first time, the `DataReportDash` has not been created.

Maybe you need to create your own `get` method for this `DataReport` form view in where to have that import statement?

dalonsoa (commented):
I think you are importing the dash app in the wrong place, but not entirely sure.

dalonsoa:
@CWestICL , if this PR is no longer relevant, just close it so it does not cause confusion.

dalonsoa:
@CWestICL , can this be closed?

PR #152, opened 2024-02-07
Data Report dash app

CWestICL:
Created a new dash app that can be accessed at http://localhost:8000/validated/data_report_dev/. This replicates the functionality of the original `data_report` page entirely within the dash app. The app layout and callback function is currently in a state where changing values in the dropdowns/date picker will immediately update the graph.

Still to be done:
- There is no error handling at the moment so if the parameters are set in a way that provides no data, the graph doesn't update
- General layout, headings etc. need to be changed
- Need to add the minimum/maximum plots

### DIFF
--- a/validated/dash_apps/finished_apps/data_report.py
+++ b/validated/dash_apps/finished_apps/data_report.py
@@ -1,0 +1,122 @@
+from datetime import datetime
+from decimal import Decimal
+
+import plotly.express as px
+from dash import dcc, html, Input, Output
+from django_plotly_dash import DjangoDash
+
+import pandas as pd
+
+from station.models import Station
+from validated.functions import dict_data_report
+from variable.models import Variable
+
+# Create a Dash app
+app = DjangoDash("DataReport")
+
+
+# Initial values
+station: Station = Station.objects.order_by("station_code")[7]
+variable: Variable = Variable.objects.order_by("variable_code")[0]
+start_time: str = datetime.strptime("2023-03-01", "%Y-%m-%d")
+end_time: str = datetime.strptime("2023-03-31", "%Y-%m-%d")
+temporality: str = "measurement"
+
+
+def plot_graph(
+    temporality: str,
+    station: Station,
+    variable: Variable,
+    start_time: str,
+    end_time: str,
+):
+    # Load data
+    data: dict = dict_data_report(
+        temporality=temporality,
+        station=station,
+        variable=variable,
+        start_time=start_time,
+        end_time=end_time,
+    )
+
+    # Create plot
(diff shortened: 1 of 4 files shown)


validated/dash_apps/finished_apps/data_report.py
```diff
@@ -0,0 +1,122 @@
...
+    Output("data_report_graph", "figure"),
+    [
+        Input("temporality_drop", "value"),
+        Input("station_drop", "value"),
+        Input("variable_drop", "value"),
+        Input("date_range_picker", "start_date"),
+        Input("date_range_picker", "end_date"),
+    ],
+)
+def update_button_click(
```
dalonsoa, line 103: I'm a bit confused about this: what button is to be clicked?

dalonsoa (approved):
This looks good to me, considering the limitations you already mention.

dalonsoa:
By the way, QA should pass. Are you running the pre-commit hooks with every commit? See https://github.com/ImperialCollegeLondon/paricia?tab=readme-ov-file#continuous-integration

validated/dash_apps/finished_apps/data_report.py
CWestICL: Oh whoops, that was a name left over from a callback function from another project that I used as a template. I forgot to rename it, I'll update that when I add some of the missing functionality

tsmbland (approved):
Looks good. FYI I put a css file in /static/styles to define things like fonts if you want to use

dalonsoa (changes requested):
In principle, this looks good but:
- When I select validation, I get the same plot as raw even when I haven't done any validation. Don't worry about this since it will be sorted after we adopt in full the new Measurement and Report models.
- Silently not showing any data when there's no data is a bit confusing. I'd suggest [showing an alert](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/alert/) in those cases with a meaningful message.

validated/dash_apps/finished_apps/data_report.py
```diff
@@ -0,0 +1,207 @@
...
+        init_start_time,
+        init_end_time,
+    )
+
+
+@app.callback(
+    Output("alert_div", "children"),
+    Input("data_report_graph", "figure"),
+)
+def clear_form(figure):
```
dalonsoa, line 201: Just to make sure I get this right, there are two functions called the same, and all works ok?

dalonsoa (approved):
This looks great. I think the alerts make for a big difference for the user.

I've noticed, however, that the alert is not triggered when you change the temporality, only when you select the station, variable and date range. Is that because there is no exception raised in that case?

Issue #108, opened 2023-07-25
Reports should show only valid options to chose


dalonsoa:
In the Data Report page, only options available for Station and Variable should be shown. Follows the same principle than #107 and #106 

<img width="279" alt="image" src="https://github.com/ImperialCollegeLondon/paricia/assets/6095790/7d46ec91-92a0-44e1-9c16-e29b6e71d770">

dalonsoa:
Already fixed, long ago.

PR #155, opened 2024-02-07
Daily validation dash app

tsmbland:
Creates a dash app to visualise data in tabular and graph form, and manually save to the Validated dataset. Replicates the original Daily Validation Javascript app (http://localhost:8000/validated/daily_validation/). 

The new app is a self-contained page accessible at http://localhost:8000/validated/daily_validation_dev/

For now, the data filters are specified in the code, although eventually we will want to include these as drop downs on the page (either through the existing django forms or new dash objects).

---

I'm keeping this as a master branch for #121, and merging PRs into this branch as I add new features.

PRs already merged:
#143 Making a start on the tables, showing the appropriate data, conditional styling
#145 Callbacks for adding new data and saving to validated
#151 Arranging tables into tabs and callback for opening the detailed view table
#154 Cosmetic improvements to the plot, and radio selector for choosing which information to show
#158 Date picker for choosing date for detail table
#159 Adds data filters to the top (replacing django form functionality)

---

Notes/remaining issues:
- Daily data and detail data doesn't seem to match up. For example, the Max of maxs. for 2023-03-14 is shown as 50.34, but looking at the detailed view for that day the highest maximum value is 17.42. This is the same in the original JS app. Am I just misunderstanding something?
- The buttons are sometimes very laggy, especially the 'Save to Validated' and 'Reset Validated' buttons which require calls to the database. (This seems to vary a lot though, and could relate to developing the code while it's running.)
- ~~Related to the above, the original app brings up loading messages while the button callbacks are running. This would be a good thing to do here given how long some of the callbacks take, but I haven't yet found an easy way to do this.~~ Done
- I'm unsure what the 'Selected' component of the data is meant to represent. Is this just to represent which measurements are selected in the tables? If so, why does Selected deviate from Measurement in the plot (see 30th March)
- The plot only updates when buttons are pressed, so selecting rows in the tables won't change the Selected dots in the plot until a button is pressed. This is the same as the JS app. 
- Why is there no data for March 31st?
- Why do days run from 5am-5am? Is this a timezone thing?
- Probably related to the above two, but opening detail view for April 1st gives 'No Rows To Show' (same in the JS app)
- ~~Fonts: I'm defining the default font at the top as a string, and then passing this as a style attribute to every text object. This works, but presumably there's a way of globally defining the default font for the app?~~ Edit: fixed using stylesheet
- Layout: I like being able to view the tables and the plot at the same time, but ideally they would both fit on the screen without having to scroll. I was having issues defining the height of individual divs so probably doing something wrong. I also don't know how to deal with potentially different screen sizes.
- There's a load of white space at the bottom which I don't know how to get rid of.
- We will need to add authentication somewhere so it only works for logged-in users.
- Filters at the top won't show an error message if invalid filters are selected and just won't update the tables/plot. However, this should be fixed when we address #108 
- In the js app the filters automatically fill in the min/max values once the station, variable and dates are selected, however I can't find where in the code this is done, so I haven't been able to replicate it here.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,7 @@ PyYAML==6.0.1
 uritemplate==4.1.1
 crispy-bootstrap4==2023.1
 django-plotly-dash==2.2.2
+dash-ag-grid==2.4.0
 
 ## Legacy dependency versions
 # asgiref==3.3.4
--- a/static/styles/dashstyle.css
+++ b/static/styles/dashstyle.css
@@ -1,0 +1,15 @@
+body {
+    font-family: Open Sans, Raleway, Dosis, Ubuntu, sans-serif;
+}
+
+.DateInput_input, .DateInput_input_1 {
+    font-size: inherit;
+    Height: 30px;
+}
+
+.Select-input{
+    height: 30px;
+}
+.Select-control{
+    height: 30px;
+}
--- a/validated/dash_apps/finished_apps/daily_validation.py
+++ b/validated/dash_apps/finished_apps/daily_validation.py
@@ -21,9 +21,9 @@ from validated.plots import create_validation_plot
 from validated.tables import create_columns_daily, create_columns_detail
 from variable.models import Variable
 
-DEFAULT_FONT = "Open Sans, Raleway, Dosis, Ubuntu, sans-serif"
-
-app = DjangoDash("DailyValidation")
+app = DjangoDash(
+    "DailyValidation", external_stylesheets=["/static/styles/dashstyle.css"]
+)
 
 # Initial filters
 STATION: Station = Station.objects.order_by("station_code")[7]
@@ -115,7 +115,6 @@ filters = html.Div(
     style={
         "display": "flex",
         "justify-content": "flex-start",
-        "font-family": DEFAULT_FONT,
         "font-size": "14px",
     },
 )
@@ -192,7 +191,6 @@ detail_date_picker = html.Div(
         "display": "inline-block",
         "width": "50%",
         "text-align": "right",
-        "font-family": DEFAULT_FONT,
     },
 )
 
@@ -225,7 +223,6 @@ status_message = html.Div(
     id="status-message",
     children=[""],
     style={
-        "font-family": DEFAULT_FONT,
         "font-size": "14px",
         "min-height": "20px",
         "padding-top": "5px",
@@ -245,14 +242,14 @@ plot_radio = dcc.RadioItems(
     ],
     value=PLOT_TYPE,
     inline=True,
-    style={"font-family": DEFAULT_FONT, "font-size": "14px"},
+    style={"font-size": "14px"},
 )
 
 # Layout
 app.layout = html.Div(
     children=[
         filters,
-        html.Button("Submit", id="submit-button"),
+        html.Button("Submit", id="submit-button", style={"margin-top": "10px"}),
         dcc.Loading(
             type="dot",
             children=html.Div(id="loading_top"),
@@ -267,8 +264,6 @@ app.layout = html.Div(
                     label="Daily Report",
                     id="tab-daily",
                     value="tab-daily",
-                    style={"font-family": DEFAULT_FONT},
-                    selected_style={"font-family": DEFAULT_FONT},
                     children=[
                         table_daily,
                     ],
@@ -278,9 +273,6 @@ app.layout = html.Div(
                     id="tab-detail",
                     value="tab-detail",
                     disabled=True,
-                    style={"font-family": DEFAULT_FONT},
-                    selected_style={"font-family": DEFAULT_FONT},
-                    disabled_style={"font-family": DEFAULT_FONT},
                     children=[
                         table_detail,
                     ],
--- a/validated/dash_apps/finished_apps/daily_validation.py
+++ b/validated/dash_apps/finished_apps/daily_validation.py
@@ -374,23 +374,29 @@ def callbacks(
     str,
     bool,
 ]:
-    """Callback for buttons adding and resetting Validated data
+    """Callbacks for daily validation app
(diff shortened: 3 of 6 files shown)


dalonsoa:
I answer here some of the remaingin quesitons:

> Daily data and detail data doesn't seem to match up. For example, the Max of maxs. for 2023-03-14 is shown as 50.34, but looking at the detailed view for that day the highest maximum value is 17.42. This is the same in the original JS app. Am I just misunderstanding something?

It might be that the functions doing the post-processing after querying the database are already doing some filtering. A way of checking this would be manually have a look at the actual data as soon as it gets our of the database, before any processing. 

> The buttons are sometimes very laggy, especially the 'Save to Validated' and 'Reset Validated' buttons which require calls to the database. (This seems to vary a lot though, and could relate to developing the code while it's running.)

The above processing is very slow and dodgy, so I'm not surprised it is slow. Hopefully, it can be greatly simplified when we use the new model.

> Related to the above, the original app brings up loading messages while the button callbacks are running. This would be a good thing to do here given how long some of the callbacks take, but I haven't yet found an easy way to do this.

No idea. Maybe @AdrianDAlessandro has a feedback on this?

> I'm unsure what the 'Selected' component of the data is meant to represent. Is this just to represent which measurements are selected in the tables? If so, why does Selected deviate from Measurement in the plot (see 30th March).

I've always wondered the same thing. In principle, that should just mean that is active (i.e., not being discarded after validation), but clearly there's something else. 

> The plot only updates when buttons are pressed, so selecting rows in the tables won't change the Selected dots in the plot until a button is pressed. This is the same as the JS app.

Sure, but is that the behavior we want? What do you think?

> Why is there no data for March 31st?

I guess that because the initial data loaded in the database, which is just some toy data used for development, does not go any further. 

> Why do days run from 5am-5am? Is this a timezone thing?

Most likely. The stations are in Peru, while you are in UK. The datetime objects in the database are timezone aware so they show you the information translated to your timezone.

> Probably related to the above two, but opening detail view for April 1st gives 'No Rows To Show' (same in the JS app)

Most likely, that is the case.

> Fonts: I'm defining the default font at the top as a string, and then passing this as a style attribute to every text object. This works, but presumably there's a way of globally defining the default font for the app?
> Layout: I like being able to view the tables and the plot at the same time, but ideally they would both fit on the screen without having to scroll. I was having issues defining the height of individual divs so probably doing something wrong. I also don't know how to deal with potentially different screen sizes.
> There's a load of white space at the bottom which I don't know how to get rid of.

No idea. @AdrianDAlessandro might know more about these L&F aspects of dash apps.

> We will need to add authentication somewhere so it only works for logged-in users.

We will deal with this as some point. Clearly it can be done (remember the Liionsden app I showed you), so let's not worry about this, yet. Open an issue to tackle it in a future iteration.

tsmbland:
> Cannot be chosen based on a dropdown with the available days, a date selector or by clicking on the relevant row?

Yeah good point. I think I'll go with a date selector as clicking rows might conflict with the checkboxes and adding buttons to each row (like the JS app) seems a bit complicated in Dash

tsmbland:
> Daily data and detail data doesn't seem to match up. For example, the Max of maxs. for 2023-03-14 is shown as 50.34, but looking at the detailed view for that day the highest maximum value is 17.42. This is the same in the original JS app. Am I just misunderstanding something?
> Why is there no data for March 31st?
> Why do days run from 5am-5am? Is this a timezone thing?

These have now mysteriously fixed themselves for me. Even for old versions of the code, and for the js app, so it's not a code related issue. 

@dalonsoa @CWestICL Could you have a look and let me know what times you see in the detail table (i.e. does it run from 5am to 5am or midnight to midnight), and do you see the high (50+) values in the detail table for 14th March??

tsmbland:
@AdrianDAlessandro One other issue which we touched on before - it seems to only work when I have one callback function. Even if I define a second empty callback function that doesn't do anything the app breaks completely and none of the buttons work. It's ok as I've just been putting everything in one callback function with loads of if statements, but it's getting really large now so would be good to break up if possible.

AdrianDAlessandro:
> @AdrianDAlessandro One other issue which we touched on before - it seems to only work when I have one callback function. Even if I define a second empty callback function that doesn't do anything the app breaks completely and none of the buttons work. It's ok as I've just been putting everything in one callback function with loads of if statements, but it's getting really large now so would be good to break up if possible.

I know you've already merged, but just confirming I added a separate callback that's completely disconnected from the rest and it worked, then I tried the [allow_duplicates flag](https://dash.plotly.com/duplicate-callback-outputs) and made a second one and it didn't work.

Turns out that Dash feature is broken in django_plotly_dash https://github.com/GibbsConsulting/django-plotly-dash/issues/481

tsmbland:
> > Daily data and detail data doesn't seem to match up. For example, the Max of maxs. for 2023-03-14 is shown as 50.34, but looking at the detailed view for that day the highest maximum value is 17.42. This is the same in the original JS app. Am I just misunderstanding something?
> 
> > Why is there no data for March 31st?
> 
> > Why do days run from 5am-5am? Is this a timezone thing?
> 
> These have now mysteriously fixed themselves for me. Even for old versions of the code, and for the js app, so it's not a code related issue.
> 

@dalonsoa It seems that this switch happens when I ran the scripts in #144. Flushing the database and reloading brings it back to 5am-5am (i.e. UK timezones), then running the scripts again brings me back to Peruvian(?) timezones (i.e. midnight-midnight). Haven't yet figured out what's going on, just an observation.

dalonsoa:
The synthetic data creation is... dodgy. It is assigning UCT timezone to every station rather than leaving those stations that have a timezone as they are. As a result, it is breaking the existing data. I'll update the synthetic data creation such that stations with a proper timezone set are left untouched.

PR #156, opened 2024-02-07
Add script for benchmarking queries

dalonsoa:
This PR adds the script to benchmark the performance of making queries with the new single `Measurement` model. Conditions for using this script are the same than for the synthetic data, already included in the README in another PR.

A summary of the findings can be found in the [wiki](https://github.com/ImperialCollegeLondon/paricia/wiki/Performance-benchmarking#query-results).


tsmbland (approved):
Looks good, and encouraging results!

Issue #142, opened 2024-01-29
Deploy proof of concept Paricia in AWS


dalonsoa:
- [x] Setup VM in AWS (Wouter)
- [x] Setup action for automatic container creation on release
- [x] Create release for Paricia
- [x] Setup deployment in the VM based on our [standard setup](https://github.com/ImperialCollegeLondon/rse-webapp-deployment)
- [x] Populate with some initial data and test/play around

ICHydro:
HI @dalonsoa, let me know if you have a preferred AWS instance type, or if not, what the minimum requirements are (number  of CPUs, RAM, and disk space), and I will set one up. If you send me ssh public keys of those that need access then I will upload them.

dalonsoa:
@ICHydro , I'm very unfamiliar with AWS instances - so far only worked with Microsoft Azure and ICT-managed VMs. I would go, for now, with a basic general purpose VM with 1 vCPU, 4 Gb RAM and 40 GB or so of hard drive, whatever its name is. For development purposes, that should be ok. Once we are happy with the result and are ready to go to production, I guess that can be scaled to something bigger.

I'll send you my SSH public key by email.

dalonsoa:
@ICHydro , do you need anything else from me to put this into movement?

dalonsoa:
An t3a.medium instance, which has 2 vCPU, 4GB RAM and 8GB of HD has been launched, with public IP is 13.40.162.45.

I will deploy the web app next week.

dalonsoa:
Paricia has been deployed in https://liionsden.rcs.ic.ac.uk/ for testing purposes.

@ICHydro , I've not been able to make it run in AWS. Well, it runs (the web app is up and running with all the bits and pieces) but I have not managed to make it accessible from outside. The only difference I can see with the setup in our testing server is that we have a tls certificate so we can use `https` rather than `http`, but otherwise it looks the same. I'll keep investigating, but at least we have a live server to try things (and break them) when the time comes.

dalonsoa:
Completed after @cc-a intervention. System live at https://data.riverflow.io/

PR #163, opened 2024-02-14
Publish Paricia as a docker container

dalonsoa:
Publishes Paricia in the GitHub package registry as a docker container, so it can be deployed anywhere else easily. 

When we're ready to go to production, this should run only when there's a new release. For now, during development, this will run anytime we make a push to the `develop` branch to allow for quickly testing changes.

Contributes to #142


tsmbland (approved):
Looks good

PR #165, opened 2024-02-14
Pull data to validate and create suspicious report and summary

### LINKED ISSUE #122: Adapt validation code and data reporting

### LINKED ISSUE #167: Pull data to validate and identify suspicious entries
Data to validate needs to be extracted from the database and then do some triage on its quality based on a number of flags (time entries too close together, too far apart, values too large or too small, etc.). 

Based on these flags, a report is put together to be used later on to highlight cells in the validation table that might require a close look.

dalonsoa:
Refactors the code that performs the pulling of validation data as well as the creation of a report of suspicious entries to be used in the validation form. 

Close #167


variable/models.py
```diff
@@ -63,6 +63,8 @@ class Variable(models.Model):
         trustworthy if the number of missing values in a given period > null_limit.
     """
 
+    NATURES = [("sum", "sum"), ("average", "average"), ("value", "value")]
```
tsmbland, line 66: Do you have any examples in mind of variables that might be of "sum" or "average" nature? Just curious what these are needed for

tsmbland (approved):
All looks good to me :)

variable/models.py
dalonsoa: You can check the variables defined in the [initial data](https://github.com/ImperialCollegeLondon/paricia/blob/develop/utilities/data/variable_variable.json). Some, like `precipitation` are cumulative (so they use `sum`); others are not cumulative, but typically provided as an average over that time or as an instantaneous value at the time of measurement. In the old system, there was no way of distinguishing between these two, but the initial refactoring done last year required this distinction.

tsmbland: Ok thanks. I understand the distinction now. I can see that this PR removes the is_cumulative field from the data in the json, but doesn't replace it with the nature field. Is this intended?

dalonsoa: `is_cumulative` is left as a property, as is it still a useful check. If you refer to the changes to the initial data, you're right I should have been a bit more careful and replace `is_cumulative=true` with `nature=sum` and `is_cumulative=false` with `nature=value` (as there's no way to distinguish `value` and `average`). I'll do so next. Thanks for pointing this.

PR #171, opened 2024-02-22
Refactor reports

### LINKED ISSUE #169: Launch the calculation of reports based on new validation data
Based on the new validated data, hourly, daily and monthly reports need to be created. This calculation can follow a similar approach currently in use, launching the processes in their own threads, sending control back to the user, as their availability is not needed immediately.

dalonsoa:
The functionality to create, save and retrieve reports is completely refactored to use the new `Reports` model and efficient, vectorised calculations. The two top level functions that will be used by other parts of Paricia are:

- `launch_report_calculations`: This will launch the calculation of the reports, calling a bunch of other functions along the way. While originally it was planned to be run in a separate thread, I believe is going to be fast enough to be run on the fly.
- `get_report_data_from_db`: Simply queries the database for the chosen report and do some minimal processing.

Fixes #169


tsmbland (approved):
This looks good, although I got a bit confused thinking about timezones while reading through this. I think it might help if the docstrings for some of the functions included details about the expected timezone of the `start_time` and `end_time` inputs (i.e. are these in the timezone of the station, UTC, or something else?), particularly when the input is a string

dalonsoa:
Unless I'm missing something, whenever there's a date as input string, this is always re-interpreted as in the time of the Station. Happy to clarify that in the docstring, but please let me know if you spot a place where I've not been consistent with this.

tsmbland:
Cool. I think it's all consistent, but might just help with readability to be more explicit in the docstrings. Your call.

dalonsoa:
@tsmbland update to the docstrings done in [3f69e96](https://github.com/ImperialCollegeLondon/paricia/pull/171/commits/3f69e96c2fced6d43b17a3823c47bd400c2b5d5d)

PR #173, opened 2024-02-22
Dash apps integrated into web pages

### LINKED ISSUE #136: Integrate dash apps into main app pages
Dash apps can be kept separate from the main pages of the django app while they are being developed, but eventually will need to be integrated into the pages (unless we decide to keep them separate). This will involve a few steps:
- Inserting the div for the dash app at the appropriate place
- Interface dash app with forms. The pages currently use forms/filters (see [here](https://github.com/ImperialCollegeLondon/paricia/blob/60a30b688977a2332d96b01828fd310b4b9e75d5/validated/forms.py#L8) and [here](https://github.com/ImperialCollegeLondon/paricia/blob/60a30b688977a2332d96b01828fd310b4b9e75d5/validated/views.py#L318)) to select the appropriate data to plot, and a submit button. We will need to set this up so that when submit is pressed it opens up the dash app using the appropriate filters from the forms.
- An alternative to the above would be to rewrite the forms to be part of the dash app

CWestICL:
The `daily_validation.html` and 'data_report.html` templates and their respective Views have been updated to display the newly created dash apps. The dev versions of these pages have been removed.

 We'll need to create a new issue to go through and remove any parts of the codebase that are now obsolete, such as the `.js` scripts for these pages.

Closes #136


tsmbland (approved):
Looks good! I've just added a tiny change which prevents a double scroll bar appearing on my screen for the validation app

dalonsoa (approved):
This looks good, but just to confirm, this breaks backwards compatibility, right? The old daily validation forms will no longer be accessible, as the templates will not use them, correct?

tsmbland:
> I believe that's correct, yeah. I'm not exactly sure how forms were being used so @tsmbland might have to confirm.

Correct. The forms are no longer accessible

PR #174, opened 2024-02-22
Set up permissions system

### LINKED ISSUE #160: Setup robust permission system
At least 3 levels of permissions at the station level for the data it contains:
    - Public (visible without login)
    - Internal (login in required)
    - Private (only visible to the owner)
 
Another potential level would be to share data with other users, but this is a potential future need
Might want to make the station metadata public while keeping data private
A user owns one or more stations but, for now, stations are owned by only one person
A user has full powers to create/delete/modify stations

tsmbland:
Using django guardian to set up an object-level permissions system. 
Close #160 

---

This is a brief overview of what I have in mind for the various objects, with checkboxes showing what I've implemented so far:

__Sensors and metadata objects (e.g. station type, sensor type, country)__
- [x] these have an owner and a permissions level (public or private)
- [x] regardless of the permissions level, only the owner can modify or delete
- [x] public objects can be used by any station, but private objects can only be used by stations with the same owner
- [ ] admins can create public or private objects, but regular users can only create private objects

__Stations__
- [x] these have an owner and a permissions level (public, private or internal)
- [x] regardless of the permissions level, only the owner can modify or delete
- [ ] only station owners can create measurements associated with the station

__Measurements__
_I'm not actually assigning any permissions to measurements. Rather, the following behaviour will be achieved by checking the permissions of the station associated with each measurement_
- [x] only the owner can modify or delete measurements
- [x] if the station is private, only the owner has view permissions
- [x] if the station is internal, all logged in users have view permissions (NB: Permissions are applied when the objects are created. Therefore, I think that users that are created after a station is created will have no permissions towards that station, which isn't ideal. I guess this could be addressed by using groups)
- [ ] if the station is public, all users (logged in or not) have view permissions 

__Variables__
- [x] these have an owner and a permissions level (public or private)
- [x] regardless of the permissions level, only the owner can modify or delete
- [x] public variables can be used by any measurement, but private variables can only be used by measurements with the same owner
- [ ] admins can create public or private variables, but regular users can only create private variables

---

The has required a number of changes to each model in the following files
- models.py: Here models are given an owner and a permissions level. I've also changed what happens when some objects are deleted (i.e. `SET_NULL` -> `CASCADE`)
- signals/handlers.py: This applies permissions to objects when they are saved
- apps.py: This registers the handlers file
- admin.py: This applies permissions to the admin page, so that only users with the appropriate permissions have view/change/delete access via the admin page

Where possible I have created base classes in the management app which the models then inherit, to reduce duplicated code.

---

This also includes #179, which added authentication to the validation app and modified the stations list to show only the stations that the user has change access for.

---

I've also added various tests. Currently I have commented out the tests relating to inactive users as these are failing, but I hope to address this in a future issue.

For some reason `tests.measurement.test_reporting.TestReporting.test_launch_reports_calculation` is now failing and will need to be fixed before we merge, but I'm not quite sure what's going on here.

---

To do list:

- [ ] #202
- [ ] #203
- [ ] #204 
- [ ] #205 
- [ ] #206 
- [ ] #207


tsmbland:
@dalonsoa This is very much a work in progress, but something isn't working properly as my permissions tests are failing and I can't figure out why. Is there something obvious that I'm missing?

measurement/signals/handlers.py
```diff
@@ -0,0 +1,40 @@
...
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from guardian.shortcuts import assign_perm
+
+from measurement.models import MeasurementBase
+
+User = get_user_model()
+
+
+@receiver(post_save, sender=MeasurementBase)
```
dalonsoa, line 11: `MeasurementBase` is an abstract model, so it does not have a table in the DB and therefore it will never be sending any signal. It should be `Measurement` - and possibly `Report`. I'm not sure if it can be applied to both by providing a list to `sender`.

measurement/signals/handlers.py
```diff
@@ -0,0 +1,40 @@
...
+def set_permissions(sender, instance, **kwargs):
+    """Set object-level permissions"."""
+    # Get permissions for model
+    delete, change, view = _get_perm_codenames(sender)
+
+    # Assign change and delete permissions for owner
+    for perm in [change, delete]:
+        assign_perm(perm, instance.owner, instance)
```
dalonsoa, line 19: `instance` is an instance of the model sending the `post_save` signal, `Measurement` or `Report`, which do not have an owner. I believe you would need to use `instance.station.owner`.

dalonsoa (commented):
I've spotted a couple of errors, but I'm not sure if that will sort everything. Let me know if it is still not working.

tsmbland:
Thanks! I've fixed the errors, however the problems remain. Looks like station-level permissions aren't being applied properly

dalonsoa:
How are you creating the station? I believe there are some ways of doing that that do not call the `post_save` signal.

tsmbland:
> How are you creating the station? I believe there are some ways of doing that that do not call the `post_save` signal.

Yes, it does look like `post_save` isn't being called properly. I'm using `Station.objects.create`

dalonsoa:
Ok, I think I've found the issue. The handlers need to be imported to be registered correctly. In [this post](https://stackoverflow.com/a/62167297/3778792), they do it in some odd way. In Liionsden, this is done by importing the handlers in the app config module:

https://github.com/ImperialCollegeLondon/Faraday-liionsden/blob/5ee795e134cd9d58d6dc9c624014c89bb5923257/management/apps.py#L9

I guess you might need to do this for both, the station and the measurements apps.

tsmbland:
Thanks! This fixed the issue

tsmbland:
@dalonsoa This isn't completely finished, but it's getting pretty large now so I think it's worth reviewing what we've got so far if you have a chance

sensor/models.py
```diff
@@ -13,11 +13,30 @@
...
 
+from django.contrib.auth import get_user_model
 from django.db import models
 from django.urls import reverse
 
+User = get_user_model()
 
-class SensorType(models.Model):
+
+class SensorBaseModel(models.Model):
```
dalonsoa, line 23: I'd call this class `PermissionBase` or something like that and put it in the [`management` app models](https://github.com/ImperialCollegeLondon/paricia/blob/develop/management/models.py). From there, use it wherever is required.

station/admin.py
```diff
@@ -0,0 +1,155 @@
...
+    PlaceBasin,
+    Region,
+    Station,
+    StationType,
+)
+
+admin.site.site_header = "Paricia Administration - Stations"
+
+
+class BaseAdmin(GuardedModelAdmin):
```
dalonsoa, line 22: The `BaseAdmin` model might also be a good candidate to go to `management.admin` and then reuse it where needed.

station/models.py
```diff
@@ -23,151 +24,130 @@
...
 BASIN_FILE_PATH = "station/basin_file/"
 
+User = get_user_model()
 
-class Country(models.Model):
-    """
-    The country that a station or region is in.
-    """
+
+class StationBaseModel(models.Model):
```
dalonsoa, line 30: See my other comment.

station/models.py
```diff
@@ -23,151 +24,130 @@
...
+    """
+    The country that a station or region is in.
+    """
+
     def get_absolute_url(self):
         return reverse("station:country_detail", kwargs={"pk": self.pk})
 
-    class Meta:
-        ordering = ("id",)
+    class Meta(StationBaseModel.Meta):
```
dalonsoa, line 61: I've never seen this. Is it really necessary to inherit from the base class meta? For some reason, I thought this was picked automatically.

station/models.py
```diff
@@ -23,151 +24,130 @@
...
+
+    PERMISSIONS_LEVELS = [
+        ("public", "Public"),
+        ("private", "Private"),
+    ]
+
+    permissions_level = models.CharField(
+        max_length=8, choices=PERMISSIONS_LEVELS, default="private"
+    )
+
```
dalonsoa, line 150: Why adding this manually instead of inheriting from the right class?

station/models.py
```diff
@@ -199,6 +179,18 @@ class Station(models.Model):
...
+    PERMISSIONS_LEVELS = [
+        ("public", "Public"),
+        ("internal", "Internal"),
+        ("private", "Private"),
+    ]
+
+    permissions_level = models.CharField(
+        max_length=8, choices=PERMISSIONS_LEVELS, default="private"
+    )
+
```
dalonsoa, line 193: Idem.

station/signals/handlers.py
```diff
@@ -0,0 +1,50 @@
...
+    # Get permissions for model
+    delete, change, view = _get_perm_codenames(sender)
+
+    # Assign view permissions for all users
+    assign_perm(view, User.objects.all(), instance)
+
+    # Assign change and delete permissions for owner
+    if instance.owner:
+        for perm in [change, delete]:
+            assign_perm(perm, instance.owner, instance)
```
dalonsoa, line 39: I mention this here, but it applies to all other handlers. 

If a permission is not assigned explicitly, does the object have it? What a re the default permissions of an object> In other words, if we create a `Basin`, for example, and not permissions are granted to the owner, what permissions they have? And everyone else? Should those permissions be removed, in addition to grant new ones to the owner?

measurement/signals/handlers.py
```diff
@@ -0,0 +1,41 @@
...
+from guardian.shortcuts import assign_perm
+
+from measurement.models import Measurement, Report
+
+User = get_user_model()
+
+
+@receiver(post_save, sender=Measurement)
+@receiver(post_save, sender=Report)
+def set_permissions(sender, instance, **kwargs):
```
dalonsoa, line 13: We might need to come up with a different plan to limit what data is accessible without assigning permissions at object level. 

Measurements (and Reports) are normally created in bulk, using [`bulk_create`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#bulk-create) as that is way, way faster than going element by element. However, using that function does not call the `save` method of the object and therefore there is no `post_save` signal. As a consequence, the permissions of those objects will not be updated. 

I'm not entirely sure what the best way forward would be, to be honest.

dalonsoa (changes requested):
This looks fantastic. It is massive, but it makes sense that it affects a lot of parts of the tool. 

I've made a couple of comments and suggestions.

station/models.py
tsmbland: The reason I didn't inherit from the base class here is that this one has an extra permission level ("internal") not found in the base class. Is there a better way of doing this?

measurement/signals/handlers.py
tsmbland: I wonder if we even need to assign permissions to measurements, as any attempt to view/change/delete a measurement could probably just check the permissions of the associated station. We could add a 'view_measurements' permission to the station model to distinguish between permissions to view the station metadata and permissions to view associated measurements.

dalonsoa: That sounds like a good plan.

tsmbland:
@dalonsoa I've done a bit more work on this. There's still more to do, but I've opened up some issues to tackle the remaining points. This is potentially close to a point where we can merge it (once the failing test is addressed), or we can leave this open and merge future PRs into this branch.

management/models.py
```diff
@@ -9,3 +9,45 @@ class User(AbstractUser):
...
+        max_length=8, choices=PERMISSIONS_LEVELS, default="private"
+    )
+
+    def set_permissions(self):
+        """Set object-level permissions."""
+
+        delete, change, view = _get_perm_codenames(self.__class__)
+
+        # Assign view permissions for all users
+        assign_perm(view, User.objects.all(), self)
```
dalonsoa, line 34: All users should have `view` access if the permission level is `Public` but not if `Private`, right?

measurement/admin.py
```diff
@@ -1,23 +1,67 @@
...
+        """Check if the user has the correct permission to view the object."""
+        if obj is not None:
+            return "view_measurements" in get_perms(request.user, obj.station)
+        return True
+
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        """Limit the list of stations available based on permssions."""
+        if db_field.name == "station":
+            kwargs["queryset"] = Station.objects.filter(owner=request.user)
+        return super().formfield_for_foreignkey(db_field, request, **kwargs)
```
dalonsoa, line 42: I'm not entirely sure what this one does. Are we filtering the list of measurements that are available in the Admin based on the stations the user is owner? And what about public measurements?

management/models.py
```diff
@@ -9,3 +9,45 @@ class User(AbstractUser):
...
+
+
+class PermissionsBase(models.Model):
+    """Base model for models that require permissions."""
+
+    owner = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, blank=True)
+
+    PERMISSIONS_LEVELS = [
+        ("public", "Public"),
+        ("private", "Private"),
```
dalonsoa, line 21: Are we missing internal here?

management/models.py
dalonsoa: And I guess we need to consider what happens with the `internal` case. Possibly via groups, as you mentioned.

management/models.py
dalonsoa: Ah, ok. I've just seen that `internal` is available only for the `Station`.

tests/permissions/test_permissions.py
```diff
@@ -0,0 +1,151 @@
...
+from django.test import TestCase
+
+from sensor.models import Sensor
+from station.models import Station, StationType
+from variable.models import Variable
+
+User = get_user_model()
+
+
+class BasePermissionsTest:
```
dalonsoa, line 11: This is very neat!

validated/dash_apps/finished_apps/daily_validation.py
```diff
@@ -239,6 +239,7 @@
 # Layout
 app.layout = html.Div(
     children=[
+        html.Div(id="stations_list", hidden=True),
```
dalonsoa, line 242: Why is this hidden.

dalonsoa (approved):
This looks fantastic. I've left a couple of comments/questions, but probably best to just merge and keep working on the remaining issues. To understand if this works for real use cases, it needs to be used, so let's move on and fix things when Wouter and his team start using the tools.

Let's have a discussion about groups, though.

dalonsoa:
Well, let's merge when the test is fixed.

measurement/admin.py
tsmbland: This filters the list of stations available to use when creating or modifying measurements in the admin page, i.e. to prevent people from creating measurements for stations they do not own.

validated/dash_apps/finished_apps/daily_validation.py
tsmbland: This is just the list of stations that comes through from the view (i.e. the stations the user has change permissions for). This doesn't need to be shown, but it's used in the callback function to populate the drop-down list, so it needs to be included in the layout

tsmbland:
The failing test was because we were calling save on an object that has already been created, which tries to create permissions that already exist.

I've suppressed this for now by giving view permissions for private objects only to the owner (which we might want to do anyway). In this case the object in question is private and ownerless, so doesn't have any permissions

However, this doesn't fix the underlying issue, so this is something we will need to fix

PR #177, opened 2024-02-23
Add endpoints for Measurement and Report

dalonsoa:
Adds end points - including url, view and serialiser, to query for Measurement and Report via the REST API. A custom queryset is used rather than filters, like in the existing views, seems it gives a higher degree of customisation.

An empty query is returned if no query parameters are included. This helps limiting the size of any query to the database via de API.

Finally, the table renderer is removed altogether as it was causing confusion and the apparent, but false, possibility of editing and creating entries via de API.


tsmbland (approved):
Looks good to me

PR #182, opened 2024-02-27
Update requirements

tsmbland:
Using pip-compile to update requirements. Doesn't seem to change anything, but probably worth doing anyway, unless things were pinned for a particular reason?


dalonsoa (approved):
I was meant to do this some time ago, so great you went ahead with it! I’m glad it works with Django 5.

Issue #170, opened 2024-02-21
Integrate validation backend with front end (plots and tables)


dalonsoa:
The validation page needs to be adapted to use the new validation logic, both pulling the data from the database (using #167) and pushing it back when done (using #169).

Issue #178, opened 2024-02-26
Integrate reporting backend with front end (plots and tables)


dalonsoa:
Use the new reporting code and model in the new report dash app.

Issue #107, opened 2023-07-25
Show only stations with data


dalonsoa:
Only stations that have data in at least one variable should be shown in the drop down.

<img width="367" alt="image" src="https://github.com/ImperialCollegeLondon/paricia/assets/6095790/4068cc1a-30a4-4eb6-ab4b-b5d0acd02269">

Issue #106, opened 2023-07-25
Only show available variables


dalonsoa:
When selecting a station in the `Validation` page, only variables available for a chosen station should be shown. 

<img width="359" alt="image" src="https://github.com/ImperialCollegeLondon/paricia/assets/6095790/b1b7e84b-257d-4475-8260-99be2a4a0d35">

PR #183, opened 2024-02-28
Modify initial data to populate Measurement model

dalonsoa:
As a result, the old individual measurement models are no longer populated with initial data. The dash apps will therefore not work as expected until #170 is implemented.


tsmbland:
Looks good. I'll build #170 off this, but maybe we should wait until that and #178 are done done before merging this.

@Sahil590 This modifies the data so that each measurement is associated with a station and a variable. This is relevant for #106 and #107, but you'll have to wait until I've done #170 which links the validation app with the new model

tsmbland (approved):

dalonsoa:
@tsmbland , agree. Let's built on top of this one the relevant issues and then merge when we are ready.

CWestICL (approved):
Lgtm!

PR #185, opened 2024-02-29
Integrate new model backend with validation frontend

### LINKED ISSUE #170: Integrate validation backend with front end (plots and tables)

tsmbland:
This makes a number of changes to the validation app to hook it up with new model, including some changes to the tables and plot to reflect the new data. I've also added some functions to reset validation, which work in the same way as before.

Should work pretty much the same as before, although I've removed the ability to add new rows to the data.

(One thing I've noticed is that a lot of the values and times are being flagged for no reason (i.e. lots of red cells). However, it wasn't like this when I started working on this so I suspect I may have corrupted the data somehow while working through this. Unfortunately I'm having a mega nightmare with Docker and struggling to reset the data so I can't test this right now, but curious how it looks for you)

Also, I'm not really sure what the 'is_active' flag is for, so we might have to change how I'm dealing with this.

Close #170

### DIFF
--- a/measurement/validation.py
+++ b/measurement/validation.py
@@ -303,6 +303,8 @@ def reset_validated_days(
 ) -> None:
     """Resets validation status for the selected data.
 
+    TODO: should this also reset any modified value, minimum or maximum entries?
+
     Args:
         station (str): Station code
         variable (str): Variable code
--- a/measurement/validation.py
+++ b/measurement/validation.py
@@ -269,14 +269,14 @@ def save_validated_entries(data: pd.DataFrame) -> None:
 
 
 def reset_validated_entries(ids: list) -> None:
-    """Resets validation status for the selected data.
+    """Resets validation and activation status for the selected data.
 
     TODO: should this also reset any modified value, minimum or maximum entries?
 
     Args:
         ids (list): List of measurement ids to reset.
     """
-    update = {"is_validated": False}
+    update = {"is_validated": False, "is_active": True}
     for _id in ids:
         Measurement.objects.filter(id=_id).update(**update)
 
@@ -301,7 +301,7 @@ def save_validated_days(data: pd.DataFrame) -> None:
 def reset_validated_days(
     station: str, variable: str, start_date: datetime.date, end_date: datetime.date
 ) -> None:
-    """Resets validation status for the selected data.
+    """Resets validation and active status for the selected data.
 
     TODO: should this also reset any modified value, minimum or maximum entries?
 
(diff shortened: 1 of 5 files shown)


dalonsoa (changes requested):
I've made a couple of suggestions but the main drawback is, as you mention, that after saving to validation a lot of entries become suspicious for no apparent reason. The time entries are specially worrying since times are not updated when saving the validation, so they should be identical to the beginning. The problem might be in my validation code and not in your code, but if you could investigate that, it would be great.

About the suggestions, I'd make the validation page just about validating not validated data OR resetting validated data. One or the other, but not both. Otherwise, there is no way to distinguish between validated entries and not validated entries.

When pressing "Save to validation" ALL entries in the range should be validated. The tick that is used to select rows (or days) should affect the `is_valid` field (therefore the label `deactivate?`. By validating them, you acknowledge that a particular entry has issues and you do not want to use it in future reports - therefore you deactivate it - but you don't throw it away or leave it in a not validated state. The previous code was very confusing because if some data was not validated, it staied there with `None` values, while what was really important was to flag them as valid entries to be used or not valid. 

In other words, this is the meaning of the fields:
- `is_validated`: indicate that an entry has been reviewed - and potentially updated.
- `is_active`: indicate that a reviewed entry (i.e. a validated one) is ok and can be used for reports.

tsmbland:
Thanks for the comments! I was misunderstanding the distinction between 'validated' and 'active', bit this makes a lot more sense now.

tsmbland:
Ok so this is mostly done now. I fixed the flagging issue by sorting the data by time before passing it to the flagging functions. I also think the `flag_time_lapse_status` function is a bit too strict as many of the remaining flagged entries are ones that differ by either 1 second above or below the expected time period which doesn't seem such a problem, so maybe we should add some leeway here?

The one major issue with the app is that, as entries remain in the table after validation, it's possible to 're-validate' data and overwrite what you've already done (e.g. if you validate a certain day in the detail table, you will overwrite that if you then validate the daily table). I think that the solution here is for tables to only show data that hasn't been validated yet, and either a separate table showing validated data or a button to switch between the two modes. But anyway, that all requires some larger changes so I'll do that in a separate PR.

dalonsoa (approved):
This looks good and works as expected, with the caveats mentioned. I've open two issues to tackle those comments in the future. Undoubtedly there will be more things to polish as we start to use the tool.

PR #189, opened 2024-03-05
Integrate new model backend with validation frontend

tsmbland:
@dalonsoa I accidentally merged #185 into this branch instead of develop, so you'll have to re-approve


measurement/validation.py
```diff
@@ -283,3 +297,26 @@ def save_validated_days(data: pd.DataFrame) -> None:
...
+        variable (str): Variable code
+        start_date (str): Start date
+        end_date (str): End date
+    """
+    Measurement.objects.filter(
+        station__station_code=station,
+        variable__variable_code=variable,
+        time__date__range=(
+            datetime.strptime(start_date, "%Y-%m-%d").date(),
+            datetime.strptime(end_date, "%Y-%m-%d").date(),
```
dalonsoa, line 320: I've just realised that these dates are naive - i.e. they don't have timezone attached. You will need to get the timezone from the station, first, and then construct a datetime object using that one. Otherwise, the entries that are considered within a particular day might be wrong.

dalonsoa (changes requested):
Sorry for requesting changes after approving it, but I've just found an error. Or potentially an error.

dalonsoa (approved):
LGTM!

PR #192, opened 2024-03-07
Delete unused apps and code

dalonsoa:
This PR addesses what we always wanted: getting rid of unused and obsolete apps and code. The PR is massive, but it is mostly about deleting things.

The main things that have been deleted are:
- Hourly, daily and monthly apps.
- Variable-specific measurement models, and related forms, views, filters and urls.
- The folder of unused apps (mainly containing pre-refactoring stuff).
- Unused scripts

Other apps and tests that still used any of the above have been adapted, but they were very few. The top menu has also been adapted to account for the new measurement models and remove the old ones.

The validation app will also be deleted once #178 is completed and the relevant code and templates are moved to the measurement app. A separate issue will be open about that.

As models have changed drastically, chances are you will want to re-build the docker container completely.


tsmbland (approved):
Looks good. This is very satisfying!

PR #197, opened 2024-03-07
Updated data report frontend

CWestICL:
Code changes to `measurement.reporting.py` and `validated.dash_apps.finished_apps.data_report.py` to reflect the changes made to the reporting back end in #171.

Closes #178

PR #200, opened 2024-03-08
Fix data ingestion

### LINKED ISSUE #199: Adding data in bulk (import or initial data loading) do not call `clean`
As a consequence, some of the manipulations done in that method do not take place. As they are important and required, appropriate steps need to be added such that that method is called on each individual element before adding them to the DB.

dalonsoa:
Data ingestion had a couple of issues. Importing was plainly not working because of an untested last minute change. That is fixed, now. And both, importing and initial data loading were not calling the clean method of the `Measurement` model, meaning that the `raw_` fields were not created. As a consequence, plotting the Raw Measurement in #197 was not showing anything.

This PR fixes these issues.

Fixes #199

### DIFF
--- a/importing/functions.py
+++ b/importing/functions.py
@@ -281,12 +281,17 @@ def save_temp_data_to_permanent(data_import_temp: DataImportTemp):
     file_path = str(BASE_DIR) + "/data/media/" + str(data_import_temp.file)
 
     all_data = construct_matrix(file_path, file_format, station)
-    must_cols = ["station_id", "variable_id", "time", "value"]
+
+    must_cols = ["station_id", "variable_id", "date", "value"]
     for table in all_data:
         cols = [
             c for c in table.columns if c in Measurement._meta.fields or c in must_cols
         ]
-        table = table[cols].dropna(axis=0, subset=must_cols)
+        table = (
+            table[cols]
+            .dropna(axis=0, subset=must_cols)
+            .rename(columns={"date": "time"})
+        )
         records = table.to_dict("records")
         variable_id = table["variable_id"].iloc[0]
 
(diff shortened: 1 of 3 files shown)


tsmbland (approved):
Looks good. Just a couple of questions/comments:
- Why `clean` for importing and `full_clean` for initial data?
- I can see that `save` is being called for the initial data but not for importing. This is fine as I've moved all the permissions over to the station level so currently there's no `post_save` signal for measurements. One thing we'll need to do though is check that the user has appropriate permissions for the stations they're uploading data for. Just need to think of the best way to do this.

CWestICL (approved):
Looks good to me!

dalonsoa:
> * Why `clean` for importing and `full_clean` for initial data?

Because `full_clean` calls the validation on individual fields and, it seems, that causes trouble for some input data that has more than 6 decimals, which is what the models accept. Without that validation, the data will just be ingested cutting the precision to 6 decimals. The previous version of the tool was not running the validation, either. I use `full_run` with the initial data simply because that data was OK and passing the validation.

> * I can see that `save` is being called for the initial data but not for importing. This is fine as I've moved all the permissions over to the station level so currently there's no `post_save` signal for measurements. One thing we'll need to do though is check that the user has appropriate permissions for the stations they're uploading data for. Just need to think of the best way to do this.

It's being called for the initial data explicitly because we are modifying the information _after_ adding the entries to the DBs and we are updating it. So it needs to be saved. It is also a very smal amount of data. When importing, we create the entries, run `clean` and then we add them to the DBs in bulk, and the data might be hundreds of thousands (possibly millions) of entries. So the sequence of events are different.

PR #210, opened 2024-03-12
Launch reports when saving or resetting validation

### LINKED ISSUE #208: Report logic not triggered when validated entries are saved
The logic for creating reports is not triggered when data is validated and the entries are saved. This means there is not data to plot when trying to generate an Hourly, Daily or Monthly report.

dalonsoa:
Launches the calculation of the reports when saving the validated data, either when saving days or when saving individual entries. It also deletes reports data in the relevant time range when resetting validation.

Fixes #208

### DIFF
--- a/measurement/validation.py
+++ b/measurement/validation.py
@@ -288,14 +288,11 @@ def reset_validated_entries(ids: list) -> None:
         ids (list): List of measurement ids to reset.
     """
     times = []
-    update = {"is_validated": False, "is_active": True}
     for _id in ids:
         current = Measurement.objects.get(id=_id)
         current.is_validated = False
         current.is_active = True
         current.save()
-        times.append(current.time)
-        current.update(**update)
         times.append(current.time)
 
     station = current.station.station_code
--- a/measurement/validation.py
+++ b/measurement/validation.py
@@ -287,9 +287,20 @@ def reset_validated_entries(ids: list) -> None:
     Args:
         ids (list): List of measurement ids to reset.
     """
+    times = []
     update = {"is_validated": False, "is_active": True}
     for _id in ids:
-        Measurement.objects.filter(id=_id).update(**update)
+        current = Measurement.objects.filter(id=_id)
+        current.update(**update)
+        times.append(current.time)
+
+    station = current.station.station_code
+    variable = current.variable.variable_code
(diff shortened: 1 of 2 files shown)


measurement/validation.py
```diff
@@ -277,33 +287,54 @@ def reset_validated_entries(ids: list) -> None:
     Args:
         ids (list): List of measurement ids to reset.
     """
+    times = []
     update = {"is_validated": False, "is_active": True}
     for _id in ids:
-        Measurement.objects.filter(id=_id).update(**update)
+        current = Measurement.objects.filter(id=_id)
+        current.update(**update)
+        times.append(current.time)
```
tsmbland, line 295: I get `AttributeError: 'QuerySet' object has no attribute 'time'`

measurement/validation.py
```diff
@@ -277,33 +287,54 @@ def reset_validated_entries(ids: list) -> None:
     Args:
         ids (list): List of measurement ids to reset.
     """
+    times = []
     update = {"is_validated": False, "is_active": True}
     for _id in ids:
-        Measurement.objects.filter(id=_id).update(**update)
+        current = Measurement.objects.filter(id=_id)
```
tsmbland, line 293:
```suggestion
        current = Measurement.objects.get(id=_id)
        current.is_validated = False
        current.is_active = True
        current.save()
```

Something like this?

tsmbland (changes requested):
Mostly looks good, but the reset validation button for the detail table isn't working for the reason stated below

measurement/validation.py
dalonsoa: Yes, this should do it. It's funny, I was not getting any error when trying the function...

tsmbland:
Ok now a test is failing. Not sure what's going on though as only the `reset_validated_entries` function was changed and that function isn't actually used anywhere in the test suite...

dalonsoa:
I don't know what's going on. I cannot reproduce it locally. It works well in my laptop. Can you try in your computer?

dalonsoa:
After merging @CWestICL PR and updating this branch, tests are passing, so no idea what's going on...

tsmbland (approved):

CWestICL (approved):
Seems to be working for me!

PR #211, opened 2024-03-12
Fix logout not working

### LINKED ISSUE #184: Logout not working
Logging out using the button in the top-right corner of the app doesn't work, and gives `Method Not Allowed: /api-auth/logout/`

dalonsoa:
From Django 5, logging out requires a CSRF token and so the logout option needs to be enclosed within a form and include the `{% csrf_token %}`. The current structure just did not allow for this to be implemented easily, so I've re-write entirely that part of the menu, giving also a bit more of flexibility. 

Fixes #184


tsmbland:
@dalonsoa Just merged the latest changes from main and the same test fail is back!

tsmbland (approved):
Otherwise, this looks good!

dalonsoa:
Ok, it now works. I'm going to merge before it fails again :P

PR #215, opened 2024-03-12
Group-based permissions

tsmbland:


management/models.py
```diff
@@ -29,10 +37,13 @@ def set_permissions(self):
...
         delete, change, view = _get_perm_codenames(self.__class__)
+        standard_group = Group.objects.get(name="Standard")
+        anonymous_user = get_anonymous_user()
 
-        # Assign view permissions for all users
+        # View permissions based on permissions level
         if self.permissions_level in ["public", "internal"]:
-            assign_perm(view, User.objects.all(), self)
+            assign_perm(view, standard_group, self)
+            assign_perm(view, anonymous_user, self)
```
dalonsoa, line 46: If it is internal, the anonymous user should not have `view` permissions, right?

tests/permissions/test_permissions.py
```diff
@@ -155,3 +152,14 @@ def test_measurement_permissions_public(self):
     def test_measurement_permissions_internal(self):
         """Test that only active users can view measurements for internal stations."""
         self.assert_perms("view_measurements", self.station_internal, True, True, False)
+
+
+class StationPermissionsTestNewUser(StationPermissionsTest):
```
dalonsoa, line 157: It seems there are actual tests missing here, right?

dalonsoa (changes requested):
I've made a couple of comments, but it looks good. The real test will come when we put this out there and they try to use it in weird ways.

management/models.py
```diff
@@ -29,10 +37,13 @@ def set_permissions(self):
         """Set object-level permissions."""
 
         delete, change, view = _get_perm_codenames(self.__class__)
+        standard_group = Group.objects.get(name="Standard")
+        anonymous_user = get_anonymous_user()
 
-        # Assign view permissions for all users
+        # View permissions based on permissions level
         if self.permissions_level in ["public", "internal"]:
```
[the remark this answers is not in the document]
tsmbland: For most of the models that's correct. However, stations do have the internal permissions level so I needed to add this here to deal with that.

management/models.py
tsmbland: So this is confusing. I've interpreted the internal status to relate to the view_measurements permission only. i.e. if a station is internal, the measurements are only visible to logged-in users, but the station metadata (which this permission relates to) is visible to everyone. Does this seem appropriate?

tests/permissions/test_permissions.py
tsmbland: No, it inherits the tests from the parent class

dalonsoa: Ah, ok. Very true. I've missed the inheritance.

management/models.py
dalonsoa: True. I forgot about the discussion we had the other day. Just to be sure, @ICHydro , is this the right interpretation?

management/models.py
```diff
@@ -1,14 +1,22 @@
...
+    All users are given staff status and added to the standard group.
+
     """
 
-    pass
+    def save(self, *args, **kwargs):
+        if self.username != "AnonymousUser":
+            self.is_staff = True
+        super().save(*args, **kwargs)
+        if self.username != "AnonymousUser":
```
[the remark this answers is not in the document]
tsmbland: I think that this doesn't work because this function is called when the anonymous user is created, but as it doesn't exist yet the get_anonymous_user function fails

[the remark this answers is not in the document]
tsmbland: I've done it a different way

CWestICL (approved):
All looks good to me! Can't see any glaring issues

management/models.py
```diff
@@ -1,14 +1,24 @@
-from django.contrib.auth.models import AbstractUser
+from django.contrib.auth.models import AbstractUser, Group
 from django.db import models
-from guardian.shortcuts import assign_perm
+from guardian.shortcuts import assign_perm, get_anonymous_user
+
+from djangomain.settings import ANONYMOUS_USER_NAME
```
dalonsoa, line 5: Get the settings from `django.conf`, so as a more robust approach to using the settings. [More info](https://docs.djangoproject.com/en/5.0/topics/settings/#using-settings-in-python-code).
```suggestion
from django.conf import settings
```

management/models.py
```diff
@@ -1,14 +1,24 @@
...
+    All users are given staff status and added to the standard group.
+
     """
 
-    pass
+    def save(self, *args, **kwargs):
+        if self.username != ANONYMOUS_USER_NAME:
+            self.is_staff = True
+        super().save(*args, **kwargs)
+        if self.username != ANONYMOUS_USER_NAME:
```
dalonsoa, line 19: And here:
```suggestion
        if self.username != settings.ANONYMOUS_USER_NAME:
            self.is_staff = True
        super().save(*args, **kwargs)
        if self.username != settings.ANONYMOUS_USER_NAME:
```

dalonsoa (approved):
I've made a suggestion on getting the annonymous user, but this looks good, otherwise.

PR #216, opened 2024-03-12
Make timelapse labelling more flexible

dalonsoa:
To avoid flagging entries on the basis of negligible differences, a +-2% tolerance has been added to the timelapse comparison.


tsmbland (approved):
Looks good

PR #217, opened 2024-03-13
CSV Download button

CWestICL:
Adds a button that downloads the currently plotted Dataframe as a `.csv`. Logic added so the CSV button is not displayed when the Data not found alert is triggered.


CWestICL:
Might try and add some error handling for if the `try` block fails. I tried to have an extra `Output` for the callback function to display an alert message but it seemed to break things elsewhere. Will look into it some more.

validated/dash_apps/finished_apps/data_report.py
```diff
@@ -192,13 +206,54 @@ def clear_form(n_clicks: int):
...
+        try:
+            file = get_report_data_from_db(
+                station=station,
+                variable=variable,
+                start_time=start_time,
+                end_time=end_time,
+                report_type=temporality,
+            ).to_csv(index=False)
+        except:
+            pass
```
dalonsoa, line 237: No bare `except` blocks. Capture the message and give some feedback to the user, so they know what went wrong.

validated/dash_apps/finished_apps/data_report.py
```diff
@@ -192,13 +206,54 @@ def clear_form(n_clicks: int):
...
+):
+    if n_clicks > 0:
+        try:
+            file = get_report_data_from_db(
+                station=station,
+                variable=variable,
+                start_time=start_time,
+                end_time=end_time,
+                report_type=temporality,
+            ).to_csv(index=False)
```
dalonsoa, line 235: The dataframe includes the station and variable id numbers, but that's not superuseful, to be honest, since it is the same for all the rows. Just drop those two columns.
```suggestion
            ).to_csv(index=False).drop(columns=["station", "variable"])
```

dalonsoa (changes requested):
It looks good but, as you say, we need some error control when trying to download the data.

tsmbland (approved):
Looks good to me!

dalonsoa (approved):
LGTM!!

PR #219, opened 2024-03-13
Variable drop down

Sahil590:
This changes the variable drop down to only display the variables in each station
for #106


Sahil590:
I found an issue when trying to filter the station_drop to only shows stations that are related to the variable selected. There is no relationship between what variables are at each station. @dalonsoa am I missing something else?

dalonsoa:
No, you're correct. You need to use the measurements as a bridge. Something like:
```python
variables = Measurements.objects.filter(station__station_code=station).values_list("variable", flat=True).distinct()

# and the opposite to get the stations

stations = Measurements.objects.filter(variable__variable_code=variable).values_list("station", flat=True).distinct()
```
Or something along these lines. I have not tested it, but it should be something along those lines.

Sahil590:
I am still a little confused in how to read the data in this app as there is no data. Would I need to create a data frame and then use the functions above to filter the dropdown?

dalonsoa:
`Measurement` is a table in the database (the same as `Variable` and `Station`), so you don't need to create that data - it should already be there.

It might be best if we tackle this together. I'm in a conference the full of the next week, but maybe @CWestICL or @tsmbland can give you a hand.

Sahil590:
This now works and only displays the variables with data

dalonsoa (commented):
This is looking good!

The reason the tests are failing is because you are not running the linters that reformat your code and regorganise imports. Please, follow [the instructions in the readme in relation to continuous integration](https://github.com/ImperialCollegeLondon/paricia?tab=readme-ov-file#continuous-integration) and then execute `pre-commit run --all-files` to ensure that the formatters are run. Then commit the changes and push. Things should now run.

Sahil590:
@dalonsoa I have done the pre-commit and all the checks have passed

dalonsoa (approved):
Looks great and works as expected! Good job.

Issue #221, opened 2024-03-13
Make object ownership compulsory


tsmbland:
Currently owner is an optional field for the models. This was required to get it to work with the initial data, however we need to make ownership compulsory so we'll have to make some changes here

tsmbland:
Technically `owner` is still an optional field in the sense that `null=True` and `blank=True`. That said, for all intents and purposes, ownership is mandatory as any new objects created through the admin will be owned by whoever created them, and this cannot be changed.

The only reason I haven't changed the `null`/`blank` options in the model is that this wouldn't work with the initial data, which doesn't have an owner specified.

@dalonsoa Is there anything I need to do for this, or is it ok to close the issue?

dalonsoa:
Let's leave it as it is, for now. It will need to be tackled when - in the future, someone - porting the existing production system to Paricia.

dalonsoa:
A solution for this could be to add a "default admin" user that gets assigned anything imported to the DB "manually", i.e. not using the forms. Something to consider.

PR #223, opened 2024-03-17
Maintain correct permissions when changing permissions level

tsmbland:


management/models.py
```diff
@@ -46,12 +46,17 @@ def set_permissions(self):
         if self.permissions_level in ["public", "internal"]:
             assign_perm(view, standard_group, self)
             assign_perm(view, anonymous_user, self)
+            remove_perm(view, self.owner, self)
```
dalonsoa, line 49: I'm not sure I get this. Why are you removing the `view` permission from the owner? Surely the owner should be able to view the object, right?

Also, would this work if there's no owner?

station/models.py
```diff
@@ -306,10 +306,15 @@ def set_permissions(self):
         if self.permissions_level == "public":
             assign_perm("view_measurements", standard_group, self)
             assign_perm("view_measurements", anonymous_user, self)
+            remove_perm("view_measurements", self.owner, self)
```
dalonsoa, line 309: As above.

station/models.py
```diff
@@ -306,10 +306,15 @@ def set_permissions(self):
...
         elif self.permissions_level == "internal":
             assign_perm("view_measurements", standard_group, self)
+            remove_perm("view_measurements", anonymous_user, self)
+            remove_perm("view_measurements", self.owner, self)
```
dalonsoa, line 313: And the same here. The owner, if it exist, should always have `view` permissions regardless of the `permission_level`.

I guess it will have `view` permissions via its membership to the `standard_group`, but that's somewhat irrelevant: if the owner were to be removed from that group, for whatever reason, they should still retain full powers over the objects they owns.

dalonsoa (changes requested):
I've a couple of comments on the `view` permission of the owner. You might convince me otherwise, but I don't think permissions of owner should rely on the permissions of the standard group for anything. 

Also, now that I think about it, probably `permission_level` should be called `visibility`, but that's matter for a different issue/PR.

tsmbland:
I get your point. My reasoning was that the owner will always inherit the permissions of the group, so there's no need to explicitly apply the permission to the owner. You do something similar in Liionsden. I couldn't envisage a scenario where a user would be removed from `standard_group`, but I guess it's not inconceivable. I realise it's confusing when looking at the code. Happy to go with your call on this.

You are correct that some of these won't work if there's no owner. I will fix this. Long term I think we should make ownership mandatory, which I plan to tackle with #221 

I agree that `visibility` is a better name than `permissions_level`. I will tackle this in a separate PR.

dalonsoa:
Let’s go to your approach. I think it is more how counter intuitive the code looks that having a real impact in how the application works.

dalonsoa (approved):
LGTM

PR #224, opened 2024-03-17
Limit object ownership to active user

### LINKED ISSUE #220: Limit ownership to the active user
Currently if you're creating an object through the admin page you can select anyone to be the owner. This may be something we want for superusers, but standard users should only be able to select themselves

tsmbland:
This limits the list of users when creating an object so that a user can only select themselves as the owner of an object they are creating. Currently there is also an option of selecting no owner, as object ownership is not compulsory, however I plan to address this in #221 

Close #220


dalonsoa (approved):
Would it make sense to automatically assign the owner of an object to whoever is creating it, without the need of selecting anything?

tsmbland:
How about this? This leaves the owner field in the admin page but greys it out so the current user is automatically selected and cannot be changed

dalonsoa (approved):
Yes, this looks like the right approach.

PR #227, opened 2024-03-18
Data Report Permissions

### LINKED ISSUE #203: Permissions for daily report app
Once the daily report app has been adjusted to use the new backend (#197), we must modify the stations list to show only the stations the user has the appropriate permissions for.

The process will be similar to #179, using the `view_measurements` permission instead of the `change_station` permission. We also need this page to work for users that aren't logged in, so we won't need the `LoginRequiredMixin`

CWestICL:
Adds authentication for the Data Report app by populating the stations dropdown according to the user's view permissions.

Closes #203

### DIFF
--- a/validated/dash_apps/finished_apps/data_report.py
+++ b/validated/dash_apps/finished_apps/data_report.py
@@ -76,6 +76,7 @@ plot = plot_graph(
 # Create layout
 app.layout = html.Div(
     children=[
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):

tsmbland (approved):
Looks good!

PR #230, opened 2024-03-19
Check station permissions in importing app

### LINKED ISSUE #206: Check station ownership when creating new measurements via import
We need to ensure that users can only create measurements for stations that they own

We're already checking this when adding measurements through the admin page, but we're currently not checking this in the importing app

tsmbland:
This adds a check to the DataImportFullCreate view to make sure the user has permissions for the station they're uploading measurements for

Close #206

### DIFF
--- a/importing/views.py
+++ b/importing/views.py
@@ -57,6 +57,13 @@ class DataImportTempCreate(generics.CreateAPIView):
     serializer_class = DataImportTempSerializer
 
     def perform_create(self, serializer):
+        # Check permissions
+        station = serializer.validated_data["import_temp"].station
+        if not self.request.user.has_perm("station.change_station", station):
+            raise PermissionDenied(
+                "Only the station owner can add measurements for this station."
+            )
+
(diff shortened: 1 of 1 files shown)


tsmbland:
We can hold off merging this for now in case it breaks something, but does this look reasonable?

importing/views.py
```diff
@@ -115,6 +116,11 @@ def perform_create(self, serializer):
         serializer.validated_data["user"] = self.request.user
 
         # Save the actual measurement data
+        station = serializer.validated_data["import_temp"].station
+        if not self.request.user.has_perm("station.change_station", station):
+            raise PermissionDenied(
+                "Only the station owner can add measurements for this station."
+            )
```
dalonsoa, line 123: This same check should be done above when uploading the temp data (line 58).

dalonsoa (changes requested):
Just a minor change, but it looks good, otherwise.

dalonsoa (approved):

PR #233, opened 2024-03-19
Fix bugs in admin classes

tsmbland:
This fixes some bugs in the admin classes which should allow superusers to add and change objects without problems.
- previously permissions were being checked by checking object ownership rather than the actual permissions attributes, which obviously won't work for superusers
- there were issues before if any objects didn't have an owner. Long term we will make object ownership mandatory, but until then this should fix those problems

There will still be issues with add permissions for non-superusers.

### DIFF
--- a/management/admin.py
+++ b/management/admin.py
@@ -1,10 +1,9 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.core.exceptions import PermissionDenied
-from django.db.models import Q
 from django.utils.translation import gettext_lazy as _
 from guardian.admin import GuardedModelAdmin
-from guardian.shortcuts import get_perms
+from guardian.shortcuts import get_objects_for_user, get_perms
 
 from .models import User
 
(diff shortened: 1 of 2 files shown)


dalonsoa (approved):
I've checked locally and now things work as expected in the admin. But let's hold on the merge until the end of the week.

I've one question not to do with this PR. Why the permissions are checked in `has_view_permission`, `has_change_permission` and `has_delete_permission` but for `has_add_permission` the membership to the Standard group is check instead? Shouldn't we check there the permissions, like everywhere else?

tsmbland:
> I've one question not to do with this PR. Why the permissions are checked in `has_view_permission`, `has_change_permission` and `has_delete_permission` but for `has_add_permission` the membership to the Standard group is check instead? Shouldn't we check there the permissions, like everywhere else?

I need to work on this. Currently, as add permissions aren't applied at the object level, I don't think add permissions exist. Superusers might have them, but standard users won't. I guess we just need to apply add permissions for all the models to the standard group then this approach should work.

PR #234, opened 2024-03-20
Make admin work for standard users

### LINKED ISSUE #229: Standard users unable to add objects in the admin page
Currently all the objects in the initial data are private, so it makes sense that standard users wouldn't be able to view them in the admin page. 

However, there appears to be no way of adding objects. You just get the message: "You don’t have permission to view or edit anything."

tsmbland:
This makes a number of changes to the permissions system, with the goal of fixing any problems for standard users.

- Apply the model level `add` permission to the standard user group for all models. This is achieved through a `post_migrate` signal. Previously, standard users were able to add objects due to the `has_add_permission` method in the admin class, but the models didn't appear in the admin homepage. This change fixes this and also allows the `has_add_permission` method to check the permissions directly.
- Use the `get_queryset` method in the admin class, which limits the objects that are visible based on view permissions. Without this, all objects would be visible even to standard users.
- Remove the `save_model` method from the admin class. This just checks that the user has correct permissions for the objects used by the object they're creating, but `formfield_for_foreignkey` already does the work by removing unsuitable options from the drop-downs
- Add `post_save` signal for PlaceBasin, which for some reason was missing before

I've tried to go through the admin site carefully as a standard user and I think everything is working now, although I may be missing something

Note: currently all the objects in the loaded data are private so won't appear to standard users. This is something we should change in the future

Close #229

### DIFF
--- a/management/models.py
+++ b/management/models.py
@@ -1,5 +1,6 @@
 from django.conf import settings
-from django.contrib.auth.models import AbstractUser, Group
+from django.contrib.auth.models import AbstractUser, Group, Permission
+from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from guardian.shortcuts import assign_perm, get_anonymous_user, remove_perm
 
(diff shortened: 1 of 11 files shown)


management/models.py
```diff
@@ -34,10 +35,10 @@ class PermissionsBase(models.Model):
         max_length=8, choices=PERMISSIONS_LEVELS, default="private"
     )
 
-    def set_permissions(self):
-        """Set object-level permissions."""
+    def set_object_permissions(self):
+        """Set object-level delete, change and view permissions."""
 
-        delete, change, view = _get_perm_codenames(self.__class__)
+        delete, change, view, add = _get_perm_codenames(self.__class__)
```
dalonsoa, line 41: The `add` permission is not used in this function. Should it be used somewhere?

dalonsoa (approved):
I have not tried to run it but it looks sensible and cleaner than the previous approach. Good work!

management/models.py
tsmbland: The `add` permission is used in the `set_model_permissions` function below this one

PR #235, opened 2024-03-20
Restrict permissions level of objects created through admin

### LINKED ISSUE #202: Restrict creation of public objects (except stations) to admins
Public objects for models such as StationType and Country can be used by any station. We must therefore make sure that public objects are only created by responsible users (i.e. admins) to prevent any accidental disruption. If a non-admin user needs an object that hasn't already been created as a public object, they can create it themselves as a private object.

The behaviour of station objects is a little different as the permissions level relates to the visibility of measurements.

tsmbland:
This ensures that standard users can only create private objects through the admin page. Superusers can still create public objects. The reason is to ensure that all public objects are vetted and maintained by admins, and aren't irresponsibly changed or deleted

The exception is stations, which still maintain the three options (private, public and internal) for all users. Here, the permissions level is a bit different as it relates to visibility of measurements, and all registered users should be allowed to create public measurements.

 Close #202

### DIFF
--- a/management/admin.py
+++ b/management/admin.py
@@ -1,9 +1,8 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
-from django.core.exceptions import PermissionDenied
 from django.utils.translation import gettext_lazy as _
 from guardian.admin import GuardedModelAdmin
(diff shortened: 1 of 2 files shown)


dalonsoa (approved):
LGTM!!

Issue #191, opened 2024-03-06
Display empty table and plot when the validation page just loads


tsmbland:
When you first load the validation app (or refresh the page), the plot will show all points as 'Active' regardless of their actual status. Any action that refreshes the plot (e.g pressing 'Submit', or switching to the 'Maximum' plot) fixes this so it displays the correct status.

EDIT:
When the page is first loaded, just display an empty table and an empty plot.

PR #237, opened 2024-03-21
Empty table and empty graph

Sahil590:
This displays and empty table and graph when the validation page is loaded. This is for #191


Sahil590:
PR 219 needs to be published first

tsmbland:
@Sahil590 You can change the base branch from `develop` to `filter` (at the top of this page), then this will show just the changes you've made in this PR. When the `filter` branch is merged it will automatically switch the base back to `develop`

PR #238, opened 2024-03-22
Revert "Empty table and empty graph"

Sahil590:
Reverts ImperialCollegeLondon/paricia#237


Sahil590:
The page will display an empty table and graph when the validation page is first loaded.

PR #239, opened 2024-03-24
Modify and test deletion behaviour

### LINKED ISSUE #207: Confirm desired behaviour when deleting objects
This is basically just to go through and check the `on_delete` options for all the model fields to make sure these are appropriate

tsmbland:
The aim of this PR was to go through and check the deletion behaviour for the various models. I have made some changes, and have added some tests.

The main deletion rules can be summarised as follows:
- For any optional foreign key fields (for example, the country field in the Station model), the deletion rule is set to `SET_NULL`, meaning that these fields will be set to null when the object is deleted
- For the owner field (which will be made mandatory in #221), the deletion rule is `CASCADE`, meaning that user-owned objects will be deleted when a user is deleted
- For some of the other mandatory fields I also use `CASCADE`. The exceptions are the station and variable fields for the Measurement model, which are set to `PROTECT`. This means that stations and variables (and, in turn, their respective owners) will be prevented from deletion if associated measurements exist. In other words, if you want to delete a station or variable, you must delete all associated measurements first. Hopefully this will prevent accidental deletion of measurements.

I haven't implemented a rubbish bin for now, as was suggested in the issue, because this would require substantial work

All of these rules are easy to change if we decide on something different.

Close #207


dalonsoa (approved):
This looks good to me and the criteria looks sensible. To be honest, with these measures in place - in particular, the PROTECT option, I'm not sure if adding a rubbish bin is necessary. People willing to perform potentially catastrophic deletions will need to delete a lot of stuff consiously.

PR #242, opened 2024-03-26
Working callback

Sahil590:


validated/dash_apps/finished_apps/daily_validation.py
```diff
@@ -556,10 +556,19 @@ def callbacks(
...
+def populate_stations_dropdown(station_codes, selected_variable):
+    stations_for_variable = (
+        Measurement.objects.filter(variable__variable_code=selected_variable)
+        .values_list("station__station_code", flat=True)
+        .distinct()
+    )
     return [
-        {"label": station_code, "value": station_code} for station_code in station_codes
+        {"label": station_code, "value": station_code}
+        for station_code in stations_for_variable
```
tsmbland, line 571:
```suggestion
        for station_code in stations_for_variable if station_code in station_codes
```

Something like this?

tsmbland: I can see you had an `if` statement like this before. Why did you remove it?

tsmbland (changes requested):
This is good, but you need to filter the output based on `station_codes` to make use of the permissions system. Added suggestion below

validated/dash_apps/finished_apps/daily_validation.py
Sahil590: I reviewed it with Diego yesterday and said there was no need to filter the output based on the station_codes. I did not know it was using the permissions system

tsmbland (approved):
Looks good!

PR #243, opened 2024-03-26
Populate date and min/max filters based on data

### LINKED ISSUE #240: Automatically set date range and min/max based on selected station and variable (validation app)
Currently the validation app is hardcoding the default date range and min/max values in the filters at the top. These were manually chosen based on the example data, however in the final app we shouldn't hardcode these values. Instead, these values should be left blank when the app loads, and update to appropriate default values when a station and variable are selected

tsmbland:
Previously we were hardcoding some of the initial filters when the page loads, and loading an initial dataset based on these filters. Starting from #219 and #242 we are moving to dynamically populating the filters based on the data. This PR continues in this vein, dynamically populating the date and min/max filters. Now that this is done, we no longer have to specify any initial filters, and we don't have to load any data until the Submit button is pressed.

I have added two callbacks:
- `set_date_range_min_max` which sets the default values for the date range and min/max filters at the top once a station and variable are selected. The default date range is chosen to cover the entire range of the data. The default min/max are also chosen to cover the entire range of the data. This means that, unless these values are manually changed, no values will be flagged for falling outside the permitted range. There may be a better criteria for this, but this will be easy to change if so.
- `set_detail_date_range` sets the range of the detail date picker once the data is loaded. 

I have also added some docstrings that were previously missing

I have realised that the stations callback in #242 won't work in cases where we have a mix of stations and variables with data. For example, if you had one station with air temperature data and another station with rainfall data, as soon as you select air temperature in the variables dropdown, the other station would become inaccessible in the stations list. Instead, I have made it so the stations dropdown should show all stations with ANY measurements (regardless of variable). This callback will run once when the page loads.

Close #240


dalonsoa (approved):
LGTM.

I think you made the right call with the stations. The alternative would be to manually add a "None" option to the station list that can be selected and recover all the possible variables, but that's overcomplicating things. This is tidier.

PR #244, opened 2024-03-26
Show only validated or not validated data in the table

tsmbland:


measurement/validation.py
```diff
@@ -33,18 +36,25 @@ def get_data_to_validate(
...
             variable__variable_code=variable,
             time__gte=start_time_,
             time__lte=end_time_,
             **extra,
         ).values()
-    ).sort_values("time")
+    )
+
+    if df.empty:
+        return pd.DataFrame()
```
dalonsoa, line 55: Just to confirm I get this right. If `df.empty`, that means it has the right column headers but there are no records. `pd.DataFrame()` is totally empty - no columns or records in it. Is this correct? If that's the case, why a totally empty dataframe is preferable?

measurement/validation.py
```diff
@@ -242,14 +255,18 @@ def generate_validation_report(
...
+    if not data.empty:
+        suspicious = flag_suspicious_data(
+            data, maximum, minimum, period, var.diff_error
+        )
+        summary = generate_daily_summary(
+            data, suspicious, period, var.null_limit, var.is_cumulative
+        )
+        granular = pd.concat([data, suspicious], axis=1)
+        return summary, granular
+    return pd.DataFrame(), pd.DataFrame()
```
dalonsoa, line 269: I think it is nicer to change the order of things here and keep most of the body of the function unindented:

```suggestion
    if data.empty:
        return pd.DataFrame(), pd.DataFrame()

    suspicious = flag_suspicious_data(
        data, maximum, minimum, period, var.diff_error
    )
    summary = generate_daily_summary(
        data, suspicious, period, var.null_limit, var.is_cumulative
        )
    granular = pd.concat([data, suspicious], axis=1)
    return summary, granular
    
```

measurement/validation.py
```diff
@@ -33,18 +36,25 @@ def get_data_to_validate(
     start_time_ = datetime.strptime(start_time, "%Y-%m-%d").replace(tzinfo=tz)
     end_time_ = datetime.strptime(end_time, "%Y-%m-%d").replace(tzinfo=tz)
     extra = {}
-    if not include_validated:
+    if only_validated:
+        extra = {"is_validated": True}
+    elif not include_validated:
         extra = {"is_validated": False}
```
dalonsoa, line 42: I think that in this new approach, we are showing either only validated data or only not validated data. It is not possible - or even desirable because of the confusion - to display all the data, validated or not. Therefore, a single flag (eg. `is_validated`) to control what should be shown is enough. That also makes the use of `extra` unnecessary since the flag can be used directly in the filter..

dalonsoa (changes requested):
Looks good - specially the removal of most of the global variables! I've just made a couple of suggestions to further simplify the retrieval of the data as we don't really need two flags anymore.

measurement/validation.py
tsmbland: Good point. `df.empty` just checks that there are no records, but in this case an empty dataframe will also have no columns. There is no difference with `pd.DataFrame()` so I can just return `df`. Not sure why I wasn't doing this to begin with

dalonsoa (approved):

Issue #241, opened 2024-03-26
Remove API access


tsmbland:
The database can currently be interacted with via the API. However, since the admin page can handle all the functionality of the API, and the API isn't yet set up to deal with permissions, we have proposed to remove access to the API for now.

All the code for the API will be kept, we are just removing access for now. This includes removing/hiding the dropdowns in the menu

Also ties in with #27

PR #245, opened 2024-03-26
Landing page Station map

### LINKED ISSUE #157: Add map to the landing page
Now that the reports and the validation are using dash/plotly from the onset, it might be pretty easy to integrate the use of maps to locate the stations easily, given that they include longitude and latitude information. 

Some examples about how this could be done are in here:
https://plotly.com/python/scatter-plots-on-maps/

## UPDATE

The map with the publicly available stations should be only in the landing page. The report and validation pages should be left as they are now.

CWestICL:
Adds a scatterplot map on the landing page displaying the geographical locations of the stations available to the user.

Closes #157


tsmbland (approved):
Looks great!

templates/home.html
```diff
@@ -40,4 +41,7 @@ <h1>Paricia</h1>
                 </tbody>
         </table>
 </div>
+<div class="{% plotly_class name='StationsMap' %} card" style="height: 100%; width: 100%">
+{% plotly_app name="StationsMap" initial_arguments=django_context ratio=1 %}
+</div>
```
dalonsoa, line 46: This is WAY more interesting for most users than the API specifications. Would move it to the top of the landing page.

djangomain/views.py
```diff
@@ -1,11 +1,27 @@
...
-    template_name = "home.html"
+    """
+    View for displaying the home page and map dash app.
+    """
+
+    def get(self, request, *args, **kwargs):
+        from .dash_apps.finished_apps import stations_map
+
+        stations = get_objects_for_user(
+            request.user, "view_measurements", klass=Station
```
dalonsoa, line 20: I'm not entirely sure but I think the correct permission here is `view_stations`. Otherwise I need to be logged in to see anything (at least with the permissions currently assigned to the stations). 

@tsmbland , could you confirm what the right permissions to check are in order to view stations metadata?

dalonsoa (approved):
It looks fantastic. I've just made a suggestion to move the map to the top of the landing page. The API definitions are not really that interesting, to be honest and - actually - I'm going to remove them, for now as part of #241 .

Sahil590 (approved):

djangomain/views.py
tsmbland: Ah yes good point. This should be `view_station`

PR #246, opened 2024-04-03
Callbacks to populate filters in report page

tsmbland:
This repeats the process carried out in #219 #242 and #243 to automatically populate the filters in the data report app based on the data. Since there's a lot of shared code between the data report and validation apps, I have moved most of the code to `filters.py` (which currently contains a lot of other stuff that needs to be deleted) 

I have also removed the `clear_form` callback, as this was causing clashes with my other callbacks, and isn't really needed anyway

### DIFF
--- a/validated/dash_apps/finished_apps/daily_validation.py
+++ b/validated/dash_apps/finished_apps/daily_validation.py
@@ -7,10 +7,8 @@ import plotly.express as px
 import plotly.graph_objects as go
 from dash import Input, Output, State, dcc, html
 from dash_ag_grid import AgGrid
-from django.db.models import Max, Min
 from django_plotly_dash import DjangoDash
 
-from measurement.models import Measurement
 from measurement.validation import (
     generate_validation_report,
     reset_validated_days,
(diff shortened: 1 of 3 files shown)


validated/dash_apps/finished_apps/daily_validation.py
```diff
@@ -7,18 +7,21 @@
...
     reset_validated_entries,
     save_validated_days,
     save_validated_entries,
 )
-from station.models import Station
+from validated.filters import (
+    populate_stations_dropdown_util,
+    populate_variable_dropdown_util,
+    set_date_range_util,
+    set_min_max_util,
```
dalonsoa, line 23: Out of curiosity, why the `_util` at the end?

If it is to avoid clashing names, it is more common to either `import something as something_`, which is convoluted when you import multiple things, or call the functions within the module the same way but starting with an underscore to indicate they are internal to the module. Something like:

```python
from filters import some_filter
def _some_filter(...):
    some_filter(...)
```

dalonsoa (approved):
LGTM¬ Just made a minor comment.

PR #252, opened 2024-04-10
Removes the API access for most models and reorganises the menu

dalonsoa:
Basically, it leaves the pages like this if NOT logged in:

image

And like this if logged in:

image

All endpoints are removed (commented) and only the importing and accessing the reports and the whole data end points is left, for now. We need to make sure that they are covered by the permission system. 

image


dalonsoa:
Sorted. For those particular pages, a relative path was being used in the menu bar rather an absolute one, so it was an easy fix.

tsmbland (approved):
Looks good!

PR #253, opened 2024-04-10
Remove validation app

dalonsoa:
Moving the relevant functionality to the measurements app. Some paths and import statements have also been changed, but otherwise the functionality should remain the same. 

To try this, the database would need to be re-done from scratch.


tsmbland (approved):
Looks good!

Issue #258, opened 2024-06-13
Add documentation


dalonsoa:
At the moment, there's no documentation whatsoever. Most functions and methods have docstrings but these are not easily accessible or rendered, definitely not for end users. There's obviously no end user documentation on how to upload data, validate it or create a new station. All of that needs to be put in place before meaningful contribution from other developers and end users is to be expected.

- [x] #259
- [x] #266
- [x] #260
- [x] #261
- [x] #262
- [x] #263
- [x] #264
- [x] #265

PR #271, opened 2024-08-15
Uses ruff instead of flake8 and isort and update pre-commit hooks

dalonsoa:
Modernize the linting and code formatting insfrastructure to match our current standard. 

`pydocstyle`-related codes have been disabled and will be re-enabled when working on the documentation (#258)

`mypy` configuration is almost identical to that of QwikApp, excluding the same rules, at least for now. `Paricia` is not heavily typed in many modules.

There are A LOT of files changed, but the meaningful changes can be found by filtering out the last commit, which is when pre-commit is run in all files and fixes to keep ruff and mypy happy are implemented.


dalonsoa:
Ok, Something I've done has broken the tests. Let's see...

dalonsoa:
It turns out that `ruff` removes unused `import` statements, which is generally good... except when they are required to register functionality. In this case, it had removed the `import XXX.signals.handlers` that several of the apps require in order register the signal handlers and assign permissions when creating objects. The undone of this change can be seen in https://github.com/ImperialCollegeLondon/paricia/pull/271/commits/7f306857fc94675d845570bc71243811307b585a

AdrianDAlessandro (approved):
LGTM!

Issue #257, opened 2024-06-13
Review permission system


dalonsoa:
At the moment, permission system is a bit erratic. It is unclear in the code what permissions control the different access levels and who can do what. For example,normal, registered users should be able to upload data, but they cannot. Also, a register user should be able to see objects with internal visibility in the map, but they cannot. In general, it is a bit obscure and should re-review (and possibly simplified, to start with).

PR #272, opened 2024-08-19
Update admin to include the formatting app

dalonsoa:
In order for both normal users and admin to see it, a few changes need to be done also to the models and the handler of the formatting app, since otherwise the current permission system (to be updated when tackling #257 ) wouldn't allow for a normal, registered user to see that app.


Sahil590 (commented):
Looks good. I can see that the formatting is available for all users. Is the app only in the admin, or is there a separate view for it?

dalonsoa:
Nope, just in the Admin. Most (all?) of the interaction with the models is done via the Admin.

Sahil590 (approved):

formatting/signals/handlers.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@receiver(post_save, sender=Association)
+@receiver(post_save, sender=Classification)
+@receiver(post_save, sender=Date)
+@receiver(post_save, sender=Delimiter)
+@receiver(post_save, sender=Extension)
+@receiver(post_save, sender=Format)
+@receiver(post_save, sender=Time)
+def set_object_permissions(sender, instance: PermissionsBase, **kwargs):
```
AdrianDAlessandro, line 27: I didn't realise you could chain the receivers together like this.

station/migrations/0011_alter_station_timezone.py
```diff
@@ -0,0 +1,633 @@
+# Generated by Django 5.1 on 2024-08-19 15:38
```
AdrianDAlessandro, line 1: I'm not sure why this migration is included in this PR

formatting/models.py
```diff
@@ -15,11 +15,12 @@
 from django.db import models
 from django.urls import reverse
 
+from management.models import PermissionsBase
 from station.models import Station
 from variable.models import Variable
 
 
-class Extension(models.Model):
+class Extension(PermissionsBase):
```
AdrianDAlessandro, line 23: This is clean. A nice way to add permissions to everything.

AdrianDAlessandro (commented):
Looks good. I'm not familiar enough with the DB structure to comment on if this is working as it should, but changes make sense.

station/migrations/0011_alter_station_timezone.py
dalonsoa: Because it hadn't happen before - and was necessary - so when I made the migrations actually needed for these changes, this one also popped up.

PR #283, opened 2024-08-23
Updates `Importing` app with the right permissions

### LINKED ISSUE #257: Review permission system

dalonsoa:
The `importing` app is in charge of ingesting data into Paricia, so very important... and it was a mess. This PR simplifies the app and fixes the permission system to allow ingestion of data by a registered user.

There are a lot of changes, but mostly are about deleting stuff and small fixes here and there. In summary:

- Ingesting data via the REST API has been removed, so `url.py`, `serializer.py` and `views.py` are gone in the `importing` app.
- The two step ingestion via `DataImportTemp` and DataImportFull` models have been removed, leaving just `DataImport` that does the work in one go.
- The correct base permission classes in models and admin implemented for the `formatting` app in https://github.com/ImperialCollegeLondon/paricia/pull/272 have been implemented also for the `importing` app.
- A couple of errors in `PermissionsBaseAdmin` had been fixed, so now the permissions actually make sense for all apps (but specially the `importing` app)!

Close #257


djangomain/urls.py
```diff
@@ -52,7 +52,7 @@
     path("api/token/refresh/", TokenRefreshView.as_view(), name="token_refresh"),
     path("api/token/verify/", TokenVerifyView.as_view(), name="token_verify"),
     path("measurement/", include("measurement.urls", namespace="measurement")),
-    path("importing/", include("importing.urls", namespace="importing")),
+    # path("importing/", include("importing.urls", namespace="importing")),
```
Sahil590, line 55: Why is this commented out

management/admin.py
```diff
@@ -75,13 +84,18 @@ def _get_queryset(db_field, user):
     """Return a queryset based on the permissions of the user.
 
     Returns queryset of public objects and objects that the user has change permisions
-    for.
+    for. For the case of `Station` objects, having the `change` pemrission is
```
Sahil590, line 87:
```suggestion
    for. For the case of `Station` objects, having the `change` permission is
```
Typo on permission

Sahil590 (approved):
Overall LGTM just some minor changes and questions

djangomain/urls.py
dalonsoa: For no reason. I've deleted it, as it was not in use, anyway.

PR #298, opened 2024-09-05
Docs formatting

### LINKED ISSUE #264: Create documentation for creating new data formats
This should cover everything within the `formatting` app.

dalonsoa:
Update the formatting app and creates the documentation explaining what it contains and how to use it.

This PR drastically strips the formatting app from everithing that was un-used or whose purpose was obscure, so chances are that some bits - in particular some model fields - will need to be re-added at some point, but is better that than having functionality that cannot be documented because its purpose is not known. 

As a result of all of these, multiple migrations add to be created as well as changing some of the initial data.

Finally, it adds a couple of features to `mkdocs` to render things nicely, and fixes a couple of issues with the permissions. 

Close #264


docs/Applications/index.md
```diff
@@ -0,0 +1,13 @@
...
+All functionality in Parcia is contained in several Django applications, each of them, in turn, including one or more database models that define that functionality.
+
+These pages describe in more detail all these models, what they are for and how they relate to each other. **All registered users can create new models for these applications via de Admin pages**.
+
+- [**Formatting**](formatting.md): Definitions of the different file formats that can be imported, including specifics around delimiters, headers etc.
+- **Variable:** Information about measured variables including units, max/min allowed values etc.
+- **Station:** Everything to do with physical stations including their location, region, ecosystem etc.
+- **Sensor:** Information on physical sensors including brand and type.
+- **Importing:** Entries are created in this app when datasets are imported, storing information on the the raw data file itself, the user, time of import etc.
+- **Measurement:** The actual time-series data is stored here when raw data files are imported. A separate model (table) exists for each variable type.
```
tsmbland, line 12:
```suggestion
- **Measurement:** The actual time-series data is stored here when raw data files are imported.
```

It's all one model now, right?

formatting/models.py
```diff
@@ -11,143 +11,248 @@
...
 
 
 class Extension(PermissionsBase):
-    """File extension."""
-
-    extension_id = models.AutoField("Id", primary_key=True)
-    value = models.CharField("Value", max_length=5)
+    """Extension of the data file.
+
+    It is mostly used to chose the tool to be employed to ingest the data. While it can
```
tsmbland, line 27:
```suggestion
    It is mostly used to choose the tool to be employed to ingest the data. While it can
```

formatting/models.py
```diff
@@ -11,143 +11,248 @@
...
+        name (CharField): Short name of the format entry.
+        description (TextField): Description of the format.
+        extension (ForeignKey): The extension of the data file.
+        delimiter (ForeignKey): The delimiter between columns in the data file. Only
+            required for text files.
+        first_row (SmallIntegerField): First row of the data, excluding any heading.
+        footer_rows (SmallIntegerField): Number of footer rows, to be ignored at the
+            end.
+        date (ForeignKey): Format for the date column. Only required for text files.
+        date_column (SmallIntegerField): Index of the date column, starting in 1.
```
tsmbland, line 181:
```suggestion
        date_column (SmallIntegerField): Index of the date column, starting at 1.
```

Also, what about `first_row`, does this start at zero or 1?

formatting/models.py
```diff
@@ -11,143 +11,248 @@
...
-    name = models.CharField("Format name", max_length=35)
-    description = models.TextField("Description", null=True)
-    location = models.CharField("Location", max_length=300, blank=True, null=True)
-    file = models.CharField(
-        "Archivo",
-        max_length=100,
-        blank=True,
-        null=True,
-        help_text="Only applies to automatic transmission",
+    first_row = models.SmallIntegerField(
```
tsmbland, line 212: Could use [`PositiveSmallIntegerField`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#positivesmallintegerfield)?

formatting/models.py
```diff
@@ -180,70 +285,183 @@ class Meta:
...
+        value_validator_text (CharField): Value validator text.
+        maximum_validator_column (SmallIntegerField): Index of the maximum value
+            validator column, starting in 1.
+        maximum_validator_text (CharField): Maximum value validator text.
+        minimum_validator_column (SmallIntegerField): Index of the minimum value
+            validator column, starting in 1.
+        minimum_validator_text (CharField): Minimum value validator text.
+        accumulate (SmallIntegerField): If set to a number of minutes, the data will be
+            accumulated over that period.
+        resolution (DecimalField): Resolution of the data. . Only used if it is to be
```
tsmbland, line 313:
```suggestion
        resolution (DecimalField): Resolution of the data. Only used if it is to be
```

management/models.py
```diff
@@ -24,15 +24,28 @@ def save(self, *args, **kwargs):
...
     ]
 
     visibility = models.CharField(
-        max_length=8, choices=VISIBILITY_LEVELS, default="private"
+        max_length=8,
+        choices=VISIBILITY_LEVELS,
+        default="private",
+        null=False,
+        blank=False,
+        help_text="Visibility level of the object. WARNING: Changing this setting will "
```
tsmbland, line 46: What would happen if someone tried to change an object to private after someone else has already referenced it in another object?

tsmbland (approved):
Looks good! Just a couple of comments/questions

management/models.py
dalonsoa: I wanted to prevent from an object to be made private after it had been referenced by another one, but didn't found a way of doing it. So, as it is now, nothing will happen: the existing relations will stay but it will not be possible to create new ones.

formatting/models.py
dalonsoa: Indeed. I didn't thought about that. That field still allows `0`, so we still need to do some validation to pick that case.

formatting/models.py
dalonsoa: Yep, it also starts in 1.

formatting/models.py
dalonsoa: I think I'm going to leave the field as it is - changing to the Positive version does not save adding validators, which would be the main reason for the change.

tsmbland: Fair enough. Just wondering - why is the counting from 1 and not zero anyway?

PR #350, opened 2024-09-25
Adds generic table and detail views

dalonsoa:
This PR adds generic views to show tables and the detail of the objects _independently_ of the admin, which is quite limiting and looks awful. Appropriate templates have been added - and old ones removed. Flexibility to add buttons to create, edit and delete objects with just a flag (and the corresponding urls and views) have also been included for future use. It also adds filtering.

This does not fix any issue but it is the building block for everything else.

An example of what things would look like has been implemented for Data Import. 

Table View:

image

Detail View:

image


Sahil590:
@dalonsoa after adding the html to the pre-commit it's broken the dash apps

dalonsoa:
Fixed. You will need to do `git pull --force` as I've removed that commit.

dalonsoa:
Eventually, yes. For now I coded the `DataImport` just as an example of using the templates. In the end, all the models will have their corresponding pages with appropriate links in the menu bar.

dalonsoa:
You can check it yourself now manuyally going to `http://localhost:8000/importing`

Sahil590 (approved):

dalonsoa:
Now the filters work as planned.

PR #351, opened 2024-09-25
:bug: Load initial data in the right order

dalonsoa:
`Delta_T` initial data should be loaded before `Station`, as the later contains references to the former.


Sahil590 (approved):
Can confirm this is now working

PR #352, opened 2024-09-26
Delta t removal

### LINKED ISSUE #329: User-definable Delta T can be removed

We should not need the user to create an object "Delta T". This is the time interval of the time series of a station, and can be calculated from the data. Also, the time interval is not necessarily constant for a (raw) time series, for example data may be available every 5 minutes for some part of the time series and 10 minutes for another part.

The system should be able to calculate Delta T from the time series, and ideally be able to deal with irregular time series (which have a varying delta T).

Sahil590:
Removes reference to delta_t. 
Closes #329


dalonsoa (changes requested):
The tests are failing since you left some `delta_t` here and there. Run the tests locally - they should pick that, and use the search tools in VSCode to pick missing bits all across the codebase.

measurement/validation.py
```diff
@@ -240,7 +240,7 @@ def generate_validation_report(
     Returns:
         A tuple with the summary report and the granular report.
     """
-    period = Station.objects.get(station_code=station).delta_t.delta_t
```
Sahil590, line 243: @dalonsoa I did remove all the references, but it's failing at this point 
image
Because I removed the lines above

Sahil590: Would we still need to keep the delta_t field in the model and make it non-editable by the user?

dalonsoa: No, you need to edit things further and work with the meaning of `delta_t`. So, that function is called in here: https://github.com/ImperialCollegeLondon/paricia/blob/7db743b69ae17255dc192fe617d3f0b762d5428a/measurement/reporting.py#L271

As you have deleted `delta_t`, period does not make sense, anymore, as it cannot be `Station` - it was meant to be a number¬ So you need to delete all references and functionality related to `period` as well.

Sahil590:
Not to sure why the check links are failing now

dalonsoa (changes requested):
Looking very good. Just a couple of minor changes.

Sahil590:
Yes, should be fixed now. It is removing quite a lot of the code though is this correct @dalonsoa

Sahil590:
Tests are failing here 
image
not sure why. I tested the `get_report_data_from_db ` function and was working, but in the tests it's coming up as empty

dalonsoa (changes requested):
I've left a couple of comments. Some more work to do, but nearly there.

measurement/validation.py
```diff
@@ -133,45 +109,19 @@ def flag_suspicious_data(
...
     Returns:
         A dataframe with the suspicious data.
     """
-    time_lapse = flag_time_lapse_status(data, period)
     value_difference = flag_value_difference(data, allowed_difference)
     value_limits = flag_value_limits(data, maximum, minimum)
-    return pd.concat([time_lapse, value_difference, value_limits], axis=1)
-
-
-def flag_suspicious_daily_count(
```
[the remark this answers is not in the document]
dalonsoa: The actual answer I had prepared was not saved. I'll write it again in a few minutes.

measurement/validation.py
```diff
@@ -133,45 +109,19 @@ def flag_suspicious_data(
...
-
-    Args:
-        data: The count of records per day.
-        period: The expected period for the measurements, in minutes.
-        null_limit: The percentage of null data allowed.
-
-    Returns:
-        A dataframe with the suspicious data.
-    """
-    expected_data_count = 24 * 60 / float(period)
```
dalonsoa, line 161: The long and nicely crafter answer I had prepared boils down to this:

We cannot calculate the expected data count anymore because we don't have the `period` but we can have an estimate of what the expected count would be by using the `mode` of the daily count (i.e. the most common value for the number of counts). It would be expected that most days would have the right number of entries and only some would deviate, so we could do:

```
expected_data_count = data.mode().iloc[0]
```

assuming that `data` is a series. If it where a dataframe, it would be `iloc[0, 0]`

So, recover this function and use the above as the `expected_data_count`, please.

dalonsoa:
> Tests are failing here not sure why. I tested the `get_report_data_from_db ` function and was working, but in the tests it's coming up as empty

I think this is failing because you removed the `hourly` report altogether. See my comment [here](https://github.com/ImperialCollegeLondon/paricia/pull/352#discussion_r1783962854). The `hourly` report should not be removed.

Sahil590:
Latest commit address the comment above of adding the `flag_suspicious_daily_count` function back and changing the `expected data count`. I changed the default value of completeness in the model, since it's no longer going to be calculated in the report.

dalonsoa (approved):

PR #353, opened 2024-09-27
:arrow_up: upgrade mkdocstrings-python version

dalonsoa:
Small upgrades of dependencies that was causing trouble when building the docs.


Sahil590 (approved):
LGTM

Issue #222, opened 2024-03-15
Implement progressive zooming in reports


dalonsoa:
Some time ago, we made a first attempt at coding a widget for the interactive visualization of the time series in javascript, with the ability to zoom in and out. We tested two approaches:

Our first widget pulled the entire time series from the database to the client side, and then vizualised the relevant time window on the client. This had the advantage that data had to be pulled in only once, and that zooming was quick (as the data was already at the client side). The disadvantage was that potentially a very large time series was pulled in (if the time series is long) and not all of it is needed (if the client never zooms out to the full time series). Some clients even struggled to keep the full time series in their memory.

In a second attempt, the client only requests the time slice that is visualised. This is quicker, but requires a request to the database for every zoom-out action. This can slow down the user experience. However, we mitigated this by making use of the aggregated (hourly, daily) data in the database. The widget calculated the resolution of the time series based on the size of the viewport (number of pixels) and select the coarsest time series needed to visualize. For example, if the visualization zoomed to 2 years of data, and the viewport is 600 pixels wide, then the daily data (730 data points) is perfectly fine. This method worked really well, as the amount of data transferred and processed per zoom action was minimized.

There are many such visualization widgets nowadays, and they may already have smart was to do this, but I thought I would document here in case it is still of relevance.

_Originally posted by @ICHydro in https://github.com/ImperialCollegeLondon/paricia/discussions/209_

PR #362, opened 2024-09-30
Improving the data loading into dash apps ⚡ 

Sahil590:
This PR aims to improve the way the dash apps handle the data of large sets of data and avoiding memory error messages that occur as a result of large data sets. 
I have implemented [plotly_resampler](https://predict-idlab.github.io/plotly-resampler/v0.10.0/getting_started/) to make the pages load faster


Sahil590:
@dalonsoa I was looking into using plotly resampler which does improve the performance quite a bit, but it does skew the data. I have attached an example of what it looks like.  [Here ](https://github.com/predict-idlab/plotly-resampler) is what the resampler does 

Without resample:
image

With resample:
image

dalonsoa:
I'm not sure what you mean by _skewing the data_. It seem the same to me, just less data points and each point a bit bigger, maybe related to the number of points aggregated, right?

The resample tool looks exactly what we were looking for to address #222 , so I would try to use it and, potentially, tweak its settings so it works well for us.

Sahil590:
Yeah, the points are just bigger in this one. I meant leaving out some data to render faster

dalonsoa:
I think this is the right call, so go ahead with it and see what happens. We can gather feedback from the users later on. I really like that it shows the window of data that has been aggregated - it is super useful for the user.

dalonsoa (changes requested):
As pointed out in the review, the code looks good, but the progressive zooming does not seem to work, as shown in the resampler repo. That's a big BUT because if it does not work, there's no really much point on using the resampler. Let's try with the new version of dash/plotly etc and see if we can make it work.

Sahil590:
@dalonsoa this should be fixed now was a small change luckily.

dalonsoa:
I'm sorry, but this still doesn't work for me, not even on a fresh installation. See the video. When zooming, the resampling stays the same as initially - or at least the legend does not update, as it does in the example of the resampled. Maybe it is being resampled, but that is not shown anywhere, which makes it harder to understand what is being shown. 

https://github.com/user-attachments/assets/f6fc164f-5c38-405f-8479-c41cf12997ca

dalonsoa (changes requested):
See my comment

Sahil590:
Looks like there are others trying to do the same with the plotly-resampler I found this, so I will give it a go but involves having the js as a static file https://github.com/predict-idlab/plotly-resampler/issues/82

Sahil590:
@dalonsoa Still can't get the progressive zooming to work I have seen some issues on the resampler repo, but there were for an older version. We can still use the plotly_resampler to render the graphs faster and set the limit of plots to 200k. In the meantime, I can try to implement other ways to see if we can get the pages to load faster

dalonsoa:
Ok, I'll give it a try tomorrow morning and see if I can find a solution. If not, we will arrange a meeting with Wouter to discuss how to best tackle this. If we are plotting 1M points but miss 80% of them in the re-sampled plot, that will be faster, but might not be what they need.

dalonsoa:
Maybe we can implement some manual resampling based on the zoom using some of [these ideas for synchronising zoom level](https://stackoverflow.com/questions/56611105/how-to-get-zoom-level-in-time-series-data-as-callback-input-in-dash). It is server side, so could have the advantage of sending less amount of data to the client.

Sahil590:
> Maybe we can implement some manual resampling based on the zoom using some of [these ideas for synchronising zoom level](https://stackoverflow.com/questions/56611105/how-to-get-zoom-level-in-time-series-data-as-callback-input-in-dash). It is server side, so could have the advantage of sending less amount of data to the client.

OK I'll have a look at this in the meantime.

dalonsoa:
@Sahil590 , I managed to get something working. Not what we wanted, but a step closer from where we were before. It seems to be impossible to update the same figure that triggered the zoom event, but I can update another one. I'll keep investigating and, in any case, push the updated code before the end of the day today.

https://github.com/user-attachments/assets/0c261eb5-a24d-47eb-9611-b55c2128b3a8

Sahil590:
This looks good, I thought it would be a bit too much to have multiple graphs, but I guess this works!

dalonsoa:
It works! I managed to do it with just one plot. What a pain... I've implemented it only for the Report app. We will update the Validation app if needed, based on the user experience, but it might be harder since the update plot function there is way more complicated.

Have a look at the code and, if you are happy with it, we merge.

https://github.com/user-attachments/assets/0c509385-46e1-4d31-b08d-e96a7b4639e3

measurement/dash_apps/data_report.py
```diff
@@ -111,26 +117,42 @@ def update_graph(
...
-        data = get_report_data_from_db(
-            station=station,
-            variable=variable,
-            start_time=start_time,
-            end_time=end_time,
-            report_type=temporality,
-        )
+        every = max(1, len(data) // MAX_POINTS)
+        resampled = data.iloc[::every]
+
```
Sahil590, line 145: When I was attempting this, I put all of this in the plots.py file, I guess that's why it was not working

measurement/reporting.py
```diff
@@ -177,6 +178,7 @@ def save_report_data(data: pd.DataFrame) -> None:
     )
 
 
+@lru_cache
```
Sahil590, line 181: What is this used for?

dalonsoa: It's used to cache calls to the decorated function, so if called with exactly the same inputs, the function is not run and just the same result is returned. Useful in this case to avoid querying the DB multiples times to get the same information.

https://docs.python.org/3/library/functools.html#functools.lru_cache

Sahil590:
@dalonsoa this works fine. I have yet to test it with some large data sets. Also, would we need to implement in the validation to improve the loading times for that as well?

dalonsoa:
In theory, we should implement it also in validation, but I'm not sure how useful it is going to be there since the main point there is the table and there won't be that much re-plotting. But yes, it should be there as well.

dalonsoa (approved):

dalonsoa:
But, maybe, let's open an issue and implement that in a separate PR, once this approach is fully tested in the Report app.

So, if you're happy with this, let's merge.

Issue #338, opened 2024-09-20
Link data import object with data via ForeignKey with a cascade relation when deleting the former


ICHydro:
**Is your feature request related to a problem or limitation? Please describe.**
It is not clear what happens when a user changes a data import by uploading a new file and reprocessing it. Are the original data from the previous file deleted?

dalonsoa:
This ties with #334 . At the moment, there's a complete disconnection - as there were in the old system, as far as I know - between the data import object and the data actually imported. Deleting the data import will not change the actual data, and re-processing the data will just delete the existing data if they overlap (same time period, variable and station).

I'd suggest we link both - the data import and the actual data - via a Foreign Key such that if a user deletes or reprocess the data import, we know exactly what data entries need to be modified. What we can do nothing about is about the existing - if any - reports, as these aggregate data that can span multiple files.

PR #364, opened 2024-10-02
Adds edit, create and delete views

### LINKED ISSUE #334: Enable the download of raw data files
The file cannot be downloaded because Django in production, by default, prevents from downloading raw data files for security reasons. They need to be served in a different way. If downloading the raw data files is a thing, I'll create an issue about that.

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/paricia/issues/332#issuecomment-2362803578_

### LINKED ISSUE #343: Extract data import as a separate view, not using the admin
The admin is very limiting in how it presents the data and how it can be dealt with. It requires heavy overriding of the underlying classes and, to be honest, it doesn't look too user friendly. 

A separate list view (and associated detail view, and add and edit forms) should be created.

dalonsoa:
Both, generic that will be used in other apps, and the concrete implementation for the data import app. Now this view is linked from the menu bar and can be accessed and used by any logged user. 

When implementing the edit and new views, that need to handle the ingestion of data files, a couple of extra bits of the code had to be updated to ensure that things behaved as they should.

This PR doesn't adress #334 , as enabling that is more convoluted that just something done in the views. 

Close #343


tsmbland (approved):
I've only had a quick glance at the code, but I've tried running it and it seems to work, with the following caveats/questions:

- It's still not really clear what format the data is supposed to be in. What columns need to be in the file? It's not obvious from the documentation, unless I'm missing something. And all the options for the Format field are in Spanish. 

- I'm a bit confused by the Reprocess data checkbox. I guess it's so you can re-trigger the import if something fails, but wouldn't this be better as a button?

- The submission date is off by an hour. I guess it's showing UTC, but I think it would be more useful to show in the local timezone

- The screenshot in the documentation shows an Owner field which is no longer in the form. I guess the person performing the import automatically becomes the owner, so no need for it in the form, but when I go to view the import object after creating it the owner field is empty.

- Does deleting the import delete the measurements from the database or just the import object?

dalonsoa:
Many thanks for the feedback. This is really useful. 

A few answers:

> * It's still not really clear what format the data is supposed to be in. What columns need to be in the file? It's not obvious from the documentation, unless I'm missing something. And all the options for the Format field are in Spanish.

It is unclear because the underlying design of the models I think is still a mess. You choose the format, but then the data processing is done by the classifications, which is something happening under the hood. The whole thing needs more clarity, and not just the documentation, but the actual workflow and the options presented to the user. I'll open an issue about it. 

> * The submission date is off by an hour. I guess it's showing UTC, but I think it would be more useful to show in the local timezone

I hate timezones. I'll check the best way forward, because I think there's something iffy with how time data is presented in the front end. I'll try to tackle this here, and if it seems deeper, I will open an issue. 

> * The screenshot in the documentation shows an Owner field which is no longer in the form. I guess the person performing the import automatically becomes the owner, so no need for it in the form, but when I go to view the import object after creating it the Owner field appears empty.

Mmmm... That's annoying. It should have been assigned automatically. I'll fix it. 

> * Does deleting the import delete the measurements from the database or just the import object?
No, it does not delete the associated measurements. We have an issue about that, #338 and my comment therein. I was hopping to connect data import and measurements this week, but I'm not sure if I'll have time.

dalonsoa:
@tsmbland , owner issue just fixed in 14f3123

dalonsoa:
@tsmbland , I've been checking the timezone issue and it is something that needs fixing across the whole of Paricia to make sense - also in the dash apps - so I'll open an issue and fix things there.

For reference, some times (like the submission time) should be presented to the user in the user's timezone, but others (like the data being ingested, or the plots in the dash apps) should be presented to the user in the original timezone the data belongs to. At the moment, everything is presented in UTC, which is the default timezone of the app, as indicated in the settings. This is confusing. More information, in https://docs.djangoproject.com/en/5.1/topics/i18n/timezones/

PR #370, opened 2024-10-03
Add all the formatting views

dalonsoa:
Related to the formatting models: extension, delimiter, date, time, formatting and classification. They are accessible from the menu bar. 

A couple of improvements related to the custom vies have also been addressed. 

Close #366

PR #376, opened 2024-10-04
Add all the variable views

### LINKED ISSUE #236: Create objects `detail` pages
If you navigate to an object in the admin page you'll see a VIEW ON SITE button in the top-right. Clicking this button gives `Server Error (500)`. Running locally you get, for example `Reverse for 'country_detail' not found. 'country_detail' is not a valid view function or pattern name.` We should fix this button, or remove it.

Seems like a problem with the `get_absolute_url` method for the models

This includes adding the relevant view pages (list, detail, edit, create and delete) to the following apps:

- [x] #343
- [x] #366
- [x] #367
- [x] #368
- [x] #369

The measurement app does not need views as they have their own dash apps to be managed and, in exceptional circumstances, via de Admin by admin who know what they are doing.

### LINKED ISSUE #369: Add views to the station app

dalonsoa:
Like #370 but for the station app.

Close #369 

This will also Close #236 , the umbrella issue, as this is the last one of the lot.


Sahil590 (approved):
LGTM

PR #384, opened 2024-10-08
Link data import objects with Measurements and Reports

### LINKED ISSUE #338: Link data import object with data via ForeignKey with a cascade relation when deleting the former

dalonsoa:
This PR links the `DataImport` objects with the actual `Measurement` entries they generate and, to some extent, to the associated `Report` objects obtained after validation. This way, when a `DataImport` object is deleted or re-processed, the related data can be easily found and removed (automatically, as they have a CASCADE relation). 

I've kept the deletion step based on time range, as some data - specially initial data or that migrated from the current system - will not have an associated `DataImport` object. Also, the deletion of the `Report` objects is not accurate as reports might combine data from multiple `DataImport`s, and not being deleted if the wrong `DataImport`, based on the mode of the data used for the calculation, is associated with that entry. So, again, the deletion based on the time range comes handy. 

Close #338


Sahil590 (approved):
LGTM

PR #386, opened 2024-10-09
Fix odd behavior of emojis in guidelines

### LINKED ISSUE #354: [BUG] The contributing guidelines look odd in the documentation
**Describe the bug**
The [CONTIRBUTING  guidelines look odd when rendered within the docs](https://imperialcollegelondon.github.io/paricia/contributing/). They are ok in GitHub. It seems that icons like 🎨 or 🐛 are rendered as images by mkdocs, upsetting the whole layout.

**To Reproduce**
Just go to the docs and have a look: https://imperialcollegelondon.github.io/paricia/contributing/

**Expected behavior**
To look similar to the file in GitHub: https://github.com/ImperialCollegeLondon/paricia/blob/main/CONTRIBUTING.md

**Screenshots**
image

dalonsoa:
Basically emojis within a text bloc are rendered as images, so they need to be in list in order to be considered just emojis. Not sure if that is the expected behaviour, but it is not a big dela in this case and it works. 

image

Close #354

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,11 @@ This project and everyone participating in it is governed by the [Paricia Code o
 
 ### Reporting Bugs
 
-This section guides you through submitting a bug report for Paricia. Following these guidelines helps maintainers and the community understand your report :pencil:, reproduce the behavior :computer: :computer:, and find related reports :mag_right:.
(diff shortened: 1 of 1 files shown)


SaranjeetKaur (approved):
Ah ok! Nice!

PR #387, opened 2024-10-10
Adds mandatory owner

### LINKED ISSUE #221: Make object ownership compulsory

dalonsoa:
Owner is now a required field in all relevant models, cannot be null or left blank in the forms - although it was already automatically assigned the current user. As a consequence of this change, all apps have migrations reflecting this change. When applying the migrations to existing databases, probably an one-off value will need to be provided for each table (i.e. each model).

To account initial data loading, which has no owner, a default owner has been added, no active. If/when required, such owner can be changed for specific objects by one admin. 

Close #221


utilities/data/management_user.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "model": "management.user",
+    "pk": 2,
+    "fields": {
+      "password": "pbkdf2_sha256$870000$gumjpzt6qPeQoSNwKzoHR1$HCGrkpdrdzAI+1yK+hxZMYSKHrrqNiFtWon8DsZWdjc=",
```
tsmbland, line 6: I'm sure it's fine, just seems a bit scary having a password in a repo...

tsmbland (approved):
Looks good!

utilities/data/management_user.json
dalonsoa: Sorted. I left a password since I thought that it would fail otherwise, but it seems to be working fine with an empty string...

PR #394, opened 2024-10-14
updating requirements files

Sahil590:
Updating all dependencies. In particular the django-plotly-dash version is updated to 4.3.2, which allows us to upgrade the base Dash version from 2.9.3 to 2.18.1.


dalonsoa (approved):
LGTM!

PR #396, opened 2024-10-16
WIP: Add local timezone detection and setting

dalonsoa:
This PR adds the detection of the local timezone of the user via a small JS snippet and that information is used to set the timezone of Django. In theory, this means that all date time information displayed and input in forms will be done in such local timezone, while the information storage in the DB is kept consistently in UTC.

- [ ] Check that django forms actually honour this setting
- [ ] Check that dash apps also honour that setting

### DIFF
--- a/djangomain/middleware.py
+++ b/djangomain/middleware.py
@@ -1,0 +1,21 @@
+import zoneinfo
+
+from django.utils import timezone
+
+
+class TimezoneMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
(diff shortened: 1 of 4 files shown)


dalonsoa:
The dash apps display the data in the DB time, which is UTC. Need to review how to enforce the local time, for consistency.

dalonsoa:
@Sahil590 , could you have a look at this, so we can merge?

Sahil590 (approved):
LGTM from what I can see. I tested it on different zones and worked as expected

PR #400, opened 2024-10-21
Improve loading times of the dash apps

dalonsoa:
This PR addresses the slow response of the dash apps by doing the following:

- Reducing to a minimum the number of queries to the `Measurement` table, which is big and will only get bigger, storing the required information in a new `variables` field within the `Station` model.
- Prevent the initial loading of the data in the reporting app, rather requiring from the user to click a button, as it is done in the Validation app. This lets the user decide on the station, variables and range of interest before plotting anything.


Sahil590 (approved):
It is better having the submit button and taking the dropdowns as state instead of as an input.  Currently testing it with larger set of data but taking a while to load it in. LGTM from what I can see.

PR #410, opened 2024-10-25
Reset validation if reporting fails

### LINKED ISSUE #383: [BUG] Even when there's an error during validation, the data appears validated
For example, there might be a problem during the report calculation and the whole thing fails. Even then, the data appears as validated and do not appear again for validation.

## Solution

Mark as validated the data after the calculation of the reports has succeeded, not before.

dalonsoa:
The report calculation needs the data to be validated, so what it is done is that if anything fails during the report calculation, the validation is reset. This is done for both the full day validation and the individual entries validation.

Close #383


Sahil590 (approved):
Looks good from what I can see

PR #411, opened 2024-10-25
Allow to reformat days without rounding to full months

### LINKED ISSUE #401: Enable the request of reports for less than a month
Even though you can choose any dates you want in the Request app, internally, full months are retrieved an plotted (if data is available). 

While probably ok, it will be useful to make this more flexible, so if a user ask for 3 days, they get 3 days, and not the whole month. 

**Describe alternatives you've considered**
Live with it.

dalonsoa:
This is useful to select exactly what you want to retrieve from the database to plot and download, which will speedup things for the user.

Close #401

### DIFF
--- a/measurement/dash_apps/data_report.py
+++ b/measurement/dash_apps/data_report.py
@@ -140,6 +140,7 @@ def update_graph(
         start_time=start_time,
         end_time=end_time,
         report_type=temporality,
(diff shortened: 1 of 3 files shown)


Sahil590 (approved):
LGTM. It works on my end

Issue #512, opened 2025-11-21
Implement API endpoint to download data


dalonsoa:
The API will be available only to registered users.

What data a user can download depends on what data stations are making visible. This is used in the data report view, for example:

https://github.com/ImperialCollegeLondon/paricia/blob/319fa80cfa1a5bb4f9a94459145fc25a896a97e6/measurement/views.py#L40-L42

This gets the stations with data that the user can view. If a user selects data for a station not in this list, they should get an access error. The [`filters` module](measurement/filters.py) contains most of the functions required to find what can be downloaded.

The information that a user will need to provide is:

- Station
- Variable
- Date range

Additionally, they will need to indicate if they want the:
- Raw data
- Validated data
- Or report data (hourly, daily or monthly)

Finally, within the above, the user will need to decide what traces to download. Typically, they will want the `value`, but other traces like `maximum`, `minimum`, `depth` or `direction` might also be queried. @ICHydro do you have a feeling of which of these values will be typically required?

The [`get_report_data_from_db` function](https://github.com/ImperialCollegeLondon/paricia/blob/319fa80cfa1a5bb4f9a94459145fc25a896a97e6/measurement/reporting.py#L189) can most likely be used to actually get the data to download once the above information is collected and validated. 

This API must be documented using the OpenAPI implemented in #511

dalonsoa:
Sorry, I'm not fully sure about your answer. Do you want to be able to download CSVs directly or JSON is fine and let the clients do whatever they want with it? For the long term plans of enabling external frontends to handle Paricia as a backend, probably the JSON version is preferred.

PR #525, opened 2025-12-02
Data download api

Sahil590:
Implements https://github.com/ImperialCollegeLondon/paricia/issues/512

This PR adds a REST API endpoint for downloading measurement data, with comprehensive documentation and validation. The endpoint allows authenticated users to download measurement data within specified date ranges, with support for different report types and data traces, while enforcing object-level permissions.

You can test this live by going to http://localhost:8000/swagger, ensure you are logged in normally or else it won't work. And you can see in the screenshot below, you can test the API from the documentation directly by selecting try it out. 

image

I also updated settings.py to add a production url.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -180,7 +180,7 @@ pip-tools==7.5.2
     # via Paricia (pyproject.toml)
 platformdirs==4.3.6
     # via virtualenv
-plotly==5.24.1
+plotly==6.5.0
     # via
     #   dash
     #   django-plotly-dash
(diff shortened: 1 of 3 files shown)


measurement/views.py
```diff
@@ -43,3 +64,243 @@ def get(self, request, *args, **kwargs):
...
+    pagination_class = MeasurementDataPagination
+
+    @property
+    def paginator(self):
+        """Return the paginator instance."""
+        if not hasattr(self, "_paginator"):
+            if self.pagination_class is None:
+                self._paginator = None
+            else:
+                self._paginator = self.pagination_class()
```
dalonsoa: As `pagination_class` is defined above as not `None`, we can just do:

```suggestion
            self._paginator = self.pagination_class()
```

PR #529, opened 2025-12-09
adding upload view

Sahil590:
I have implemented a new API view that allows the upload of data using a raw file

### DIFF
--- a/importing/serializers.py
+++ b/importing/serializers.py
@@ -1,40 +1,28 @@
 from rest_framework import serializers
 
-from formatting.models import Format
 from station.models import Station
 
 from .models import DataImport
 
 
-class DataImportUploadRequestSerializer(serializers.Serializer):
+class DataImportUploadRequestSerializer(serializers.ModelSerializer):
     """Serializer for data import upload request."""
 
     station = serializers.SlugRelatedField(
         slug_field="station_code",
         queryset=Station.objects.all(),
-        help_text="Station code",
     )
-    format = serializers.PrimaryKeyRelatedField(
-        queryset=Format.objects.all(),
-        help_text="Format ID",
-    )
-    visibility = serializers.ChoiceField(
-        choices=["public", "private"],
-        default="private",
-        help_text="Visibility level",
-    )
-    reprocess = serializers.BooleanField(
-        default=False,
-        help_text="Reprocess data after import",
-    )
-    observations = serializers.CharField(
-        required=False,
-        allow_blank=True,
-        help_text="Additional observations",
-    )
-    rawfile = serializers.FileField(
-        help_text="Data file to upload",
-    )
+
+    class Meta:
+        model = DataImport
+        fields = [
+            "station",
+            "format",
+            "visibility",
+            "reprocess",
+            "observations",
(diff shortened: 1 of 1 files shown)


Sahil590:
@dalonsoa How would we want the user to specify the format of the data, currently I have implemented it to use the format ID which was auto generated. Should I use the name and would they all be unique enough to work?

dalonsoa:
Yeah, I'd use the format ID, which is unique. At this stage, the user still needs to check the web page to learn about the format they want to use. In the future, we might want to get that info with an API endpoint, not not now.

importing/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+        help_text="Reprocess data after import",
+    )
+    observations = serializers.CharField(
+        required=False,
+        allow_blank=True,
+        help_text="Additional observations",
+    )
+    rawfile = serializers.FileField(
+        help_text="Data file to upload",
+    )
```
dalonsoa: I think that, in this case, you should be able to use a `ModelSerializer`, as you do below, just with other fields:


```suggestion
class DataImportUploadRequestSerializer(serializers.ModelSerializer):
    """Serializer for data import upload request."""

    class Meta:
        model = DataImport
        fields = ["station", "format", "visibility", "reprocess", "observations", "rawfile"]
```

importing/views.py
```diff
@@ -45,3 +64,157 @@ class DataImportDeleteView(CustomDeleteView):
...
+                station=station,
+                format=validated_data["format"],
+                visibility=validated_data["visibility"],
+                reprocess=validated_data["reprocess"],
+                observations=validated_data.get("observations", ""),
+                rawfile=validated_data["rawfile"],
+                owner=request.user,
+            )
+        except Exception:
+            logging.exception("Error creating data import")
```
dalonsoa: In the logs, I would add the actual error from the exception, to be able to fix it, if needed.

```suggestion
        except Exception as e:
            logging.exception(f"Error creating data import: {e}")
```
Or something along these lines.

importing/views.py
```diff
@@ -45,3 +64,157 @@ class DataImportDeleteView(CustomDeleteView):
...
+                    )
+                ],
+            ),
+            400: OpenApiResponse(description="Invalid request parameters or file"),
+            403: OpenApiResponse(
+                description="User does not have permission to upload data."
+            ),
+            404: OpenApiResponse(description="Station or format not found"),
+        },
+        tags=["data-import"],
```
dalonsoa: I think you called this `importing` somewhere else, so we should use the same thing to make it appear in the right place in the API documentation. 
```suggestion
        tags=["importing"],
```

This makes me wonder why in measurements we have `download` and `download{format}`. There should be only one, right?

Image

dalonsoa (changes requested):
I've a couple of comments, but this looks really good.

importing/serializers.py
Sahil590: @dalonsoa I would keep this as is for now, since we want to specify the `station_code` rather than the `station_id`. I am happy to change it if you would rather specify the ID

PR #533, opened 2025-12-17
Data ingestion

### LINKED ISSUE #514: Implement API to query data ingestion
Data ingestion is done in the background, so there has to be a way for a register user to query the state of the ingestion. This endpoint should:

- Provide a list of `DataImport` objects for the user, indicating their [import status](https://github.com/ImperialCollegeLondon/paricia/blob/319fa80cfa1a5bb4f9a94459145fc25a896a97e6/importing/models.py#L46).
- If a specific PK is provided, and assuming it is owned by the user, the API should return all the details of the data import, specially the log in case there has been any issue with the data ingestion.

Sahil590:
Fixes #514 
I have implemented a new GET endpoint to importing that allows users to query data that has been ingested and request details for that data if they are the owner. 
If you test it using the swagger UI, then you can only execute a standard GET request. 
If you try it in the browser will the URL below to query the PK 

http://localhost:8000/importing/api/dataingestion/?data_import_id=1

### DIFF
--- a/importing/views.py
+++ b/importing/views.py
@@ -2,7 +2,9 @@ import logging
 
 from drf_spectacular.utils import (
     OpenApiExample,
+    OpenApiParameter,
     OpenApiResponse,
+    OpenApiTypes,
     extend_schema,
 )
 from guardian.shortcuts import get_objects_for_user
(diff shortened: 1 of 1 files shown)


importing/views.py
```diff
@@ -216,5 +217,76 @@ def post(self, request):
...
+        # If a PK is provided, fetch the object and enforce owner-only access.
+        try:
+            data_import = DataImport.objects.get(pk=data_import_id)
+        except DataImport.DoesNotExist:
+            return Response(
+                {"detail": "Data import not found."}, status=status.HTTP_404_NOT_FOUND
+            )
+
+        # Only the owner may request detailed info by PK
+        if getattr(data_import, "owner", None) != request.user:
```
dalonsoa: Why not simply:

```suggestion
        if data_import.owner != request.user:
```

Issue #547, opened 2026-01-14
Add user page to manage credentials


dalonsoa:
A user page should be added to Paricia where a user can manage not only their Paricia credentials - eg. update their password or name - but also manage their Thingsboard credentials (username and password), so they don't need to be input every time a user wants to pull data. Storing it within Paricia also facilitates scheduled pulling of the dat, although for that kind of things it will be best to use an access token or something like that. 

The Thingsboard password field should not show the password by default, but only showing it by request.

dalonsoa:
Implemented in https://github.com/ImperialCollegeLondon/paricia/pull/552

PR #552, opened 2026-01-22
Account page

Sahil590:
For #547 
I have added a new user page where a logged-in user can change there account details such as the name, email, and where they can set the things board username, password and API key.

### DIFF
--- a/management/views.py
+++ b/management/views.py
@@ -39,25 +39,12 @@ class UserProfileView(LoginRequiredMixin, UpdateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context.setdefault("profile_form", self.get_form())
+        context.setdefault("profile_form", context.get("form") or self.get_form())
         context.setdefault(
             "thingsboard_form",
             ThingsboardCredentialsForm(instance=self.get_thingsboard_credentials()),
         )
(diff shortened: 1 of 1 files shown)


management/admin.py
```diff
@@ -81,6 +81,11 @@ def formfield_for_choice_field(self, db_field, request, **kwargs):
         return super().formfield_for_choice_field(db_field, request, **kwargs)
 
 
+class ThingsboardCredentialsInline(admin.StackedInline):
+    model = ThingsboardCredentials
+    can_delete = False
```
dalonsoa: Where is this used, if used at all?

dalonsoa (approved):
This looks really good! I have one question, but no blocker, so approving.

I still get the form populated with my Paricia username and password, regardless of what you did, but I think that is a lost battle with the browser. Check https://dev.to/gera2ld/how-to-disable-autofill-for-real-1411 , but ignore the suggested solution. It's too hacky!

So all good!

Sahil590:
That's annoying, I thought populating the fields with an empty string would work. Must be due to the broswer becuase on other websites i do get it autofilled when its not even the login page

dalonsoa:
Edge, based on Chrome, so probably similar behaviour.

PR #570, opened 2026-02-20
add create view for thingsboard map

### LINKED ISSUE #560: Implement view(s) to pull data from Thingsboard
This should be a separate view within `Import` that will offer the user the possibility to pull data from Thingsboard into Paricia. They will select the `ThingsboardImportMaps` to use for the import (see #543) and then submit the request. For now, this would be a dummy submission and nothing will happen. 

The `Thingsboard Import` entry in the menu should only appear if the user has `Thingsboard` credentials setup in their `Profile` (see #552).

Sahil590:
Fixes #560 
Added relevant views for manipulating the `ThingsboardImportMap`. I will add the view to create an import when I get the logic for the API working.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -15,7 +15,7 @@ attrs==25.4.0
     #   referencing
 blinker==1.9.0
     # via flask
-build==1.2.2.post1
+build==1.4.0
     # via pip-tools
 certifi==2024.8.30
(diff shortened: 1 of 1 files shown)


dalonsoa (changes requested):
I've a couple of requests, but this looks really good and works as expected.

PR #571, opened 2026-02-23
Add data origin

### LINKED ISSUE #562: Add import origin to the Data Import
The data imported into Paricia now can have different origins, not only CSV/TSV files. This should be expanded to consider two other cases: API and Thingsboard. 

To make this more modular, a `ImportOrigin` model should be added with a single field, `name`. Three objects should be automatically created in a data migration: `file`, `API` and `Thingsboard`. As this is something that very rarely will change, there's no need to create dedicated views to create, update, list and delete these objects, but it will be enough to managing them via the Admin. 

Once this model is created, the Data Import model (and any relevant view) should be updated to include the above, defaulting to `file` for any existing data.

dalonsoa:
Add an `ImportOrigin` model to declare the origin of the data being imported, and use it in the `DataImport` model. A data migration is included, pre-populating the database with `file`, `api` and `Thingsboard` origins. The views and serializers, as well as the tests, are updated to consider this new field. 

The list of data imports now look something similar to this:

image

Strictly speaking, when using the API, we also upload a file, so it might be better to indicate `web`, `api` and `thingsboard`, but let's leave it as it is for now. 

Close #562

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -151,7 +151,7 @@ markupsafe==3.0.2
     #   werkzeug
 model-bakery==1.20.1
     # via Paricia (pyproject.toml)
-mypy==1.13.0
+mypy==1.19.1
     # via Paricia (pyproject.toml)
 mypy-extensions==1.1.0
     # via mypy
--- a/requirements-doc.txt
+++ b/requirements-doc.txt
@@ -134,7 +134,7 @@ jsonschema-specifications==2025.9.1
     # via jsonschema
 lxml==5.3.0
     # via mkdocs-caption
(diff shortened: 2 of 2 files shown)


Sahil590 (approved):
Everything looks good. I have one nitpick, the migrations could be combined into one file rather than 3. But will approve otherwise.

dalonsoa:
> Everything looks good. I have one nitpick, the migrations could be combined into one file rather than 3. But will approve otherwise.

The second one is a data migration, not a model migration, so I rather keep that one separated. And as that one is the middle one, the other two cannot me merged, either.

PR #586, opened 2026-03-10
561 thingsboard logic

Sahil590:
I have made some minor changes to naming of the `tb_variable` and `tb_device_name` to make it easier to differentiate the 2. 

I have also implemented the Logic that will be used to fetch data from `thingsboard`. I added two new functions to the `Importing/Utilts.py` one is `retrieve_thingsboard_device_id` this will take the `tb_device_name` stored in `DataImportMap` and make a request through the API to retrieve the device ID as this is needed when making a request for the data. 

The next function `retrieve_thingsboard_data` set up the request to API for getting data from thingsbaord when given the device ID, customer ID, start and end dates. 

To make use of the functions a new page was made that can be accessed from the import data page. It is a form that takes the `dataimportmap`, format, start date and end date and is passed on as args for the `retrieve_thingsboard_data` function. After the data is retrieved the entry is made into the `DataImport` model and is submitted to be imported.

### DIFF
--- a/importing/utils.py
+++ b/importing/utils.py
@@ -30,10 +30,8 @@ def retrieve_thingsboard_device_id(
     """Retrieve ThingsBoard device ID for a given device name."""
     api_headers = {"X-Authorization": f"Bearer {token}", "Accept": "application/json"}
 
-    if settings.TB_CUSTOMER_DEVICES_URL is None:
-        raise Exception("TB_CUSTOMER_DEVICES_URL environment variable is not set.")
-    else:
-        devices_url = settings.TB_CUSTOMER_DEVICES_URL.format(customer_id=customer_id)
+    assert settings.TB_CUSTOMER_DEVICES_URL is not None
+    devices_url = settings.TB_CUSTOMER_DEVICES_URL.format(customer_id=customer_id)
     logger.debug(f"Using customer endpoint (customer ID: {customer_id})")
 
     r = requests.get(devices_url, headers=api_headers)
@@ -76,10 +74,8 @@ def retrieve_thingsboard_data(
     logger.debug(
         f"Retrieving ThingsBoard data for device {tb_device_id}, variable {variable}"
     )
-    if settings.TB_TIMESERIES_URL is None:
-        raise Exception("TB_TIMESERIES_URL environment variable is not set.")
-    else:
-        url = settings.TB_TIMESERIES_URL.format(tb_device_id=tb_device_id)
+    assert settings.TB_TIMESERIES_URL is not None
+    url = settings.TB_TIMESERIES_URL.format(tb_device_id=tb_device_id)
     headers = {"X-Authorization": f"Bearer {token}"}
     params: dict[str, str | int] = {
         "interval": 60000,


Sahil590:
@dalonsoa this is more or less done still need to iron out the rough edges but wanted to get some feedback before tests.

importing/utils.py
```diff
@@ -1,3 +1,12 @@
+import logging
+
+import requests
+
+from djangomain.settings import settings
```
dalonsoa: For using the Django settings you should do:


```suggestion
from django.conf import settings
```

Have a look at [the docs](https://docs.djangoproject.com/en/6.0/topics/settings/#using-settings-in-python-code) to find out more about it.

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
...
     for origin in origins:
         ImportOrigin.objects.filter(origin=origin).delete()
+
+
+def retrieve_thingsboard_device_id(token, customer_id, device_name):
+    """Retrieve ThingsBoard device ID for a given device name."""
+    api_headers = {"X-Authorization": f"Bearer {token}", "Accept": "application/json"}
+
+    devices_url = f"https://{settings.TB_HOST}/api/customer/{customer_id}/devices?pageSize=10000&page=0"
+    print(f"Using customer endpoint (customer ID: {customer_id})")
```
dalonsoa: Use the logger, not `print`.

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
...
+
```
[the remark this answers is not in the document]
dalonsoa: Exactly :) Weird to agree with Copilot...

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
...
+    r = requests.get(devices_url, headers=api_headers)
+
```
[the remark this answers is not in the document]
dalonsoa: Also agree with this one.

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
...
+    meta = r.json().get("data", [])
+
+    # Filter to only the device we need
+    device_ids = [x["id"]["id"] for x in meta if x.get("name") == device_name]
+
+    if len(device_ids) == 0:
+        raise Exception(f"Device '{device_name}' not found!")
+
+    return device_ids[0]
```
dalonsoa, line 57: I assume there should be one device only, right? I'd suggest you log a warning if there happen to be more than one.

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
     origins = ["file", "api", "Thingsboard"]
     for origin in origins:
         ImportOrigin.objects.filter(origin=origin).delete()
+
+
+def retrieve_thingsboard_device_id(token, customer_id, device_name):
```
dalonsoa: This arguments need type annotations and be added to the docstring.

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
...
+    device_ids = [x["id"]["id"] for x in meta if x.get("name") == device_name]
+
+    if len(device_ids) == 0:
+        raise Exception(f"Device '{device_name}' not found!")
+
+    return device_ids[0]
+
+
+def retrieve_thingsboard_data(
+    token, customer_id, tb_device_name, variable, start_ts, end_ts
```
dalonsoa: This arguments need type annotations and be added to the docstring.

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
...
+    Saves the response to a JSON file in the media directory.
+    """
+
+    tb_device_id = retrieve_thingsboard_device_id(token, customer_id, tb_device_name)
+    print(f"Retrieving ThingsBoard data for device {tb_device_id}, variable {variable}")
+    url = (
+        f"https://{settings.TB_HOST}/api/plugins/telemetry/DEVICE/{tb_device_id}/values/timeseries"
+        f"?interval=60000&limit=10000&agg=NONE"
+        f"&keys={variable}&startTs={start_ts}&endTs={end_ts}"
+    )
```
dalonsoa: Rather than hiding this complex URLs here (same with the previous function) I would extract them as functions or as constants defined in the settings that can be then updated where needed. So, in this case, in the settings I'd put:

```python
    TB_TIMESERIES_URL = ( 
        f"https://{TB_HOST}/api/plugins/telemetry/DEVICE/{{tb_device_id}}/values/timeseries"
        f"?interval=60000&limit=10000&agg=NONE"
        f"&keys={{variable}}&startTs={{start_ts}}&endTs={{end_ts}}"
    )
```
Not the double {{}} in the variables that are not to be replaced at that time. Only the `TB_HOST` is replaced in the definition.

And then, in here, you do:

```python
url = settings.TB_TIMESERIES_URL.format(
    tb_device_id=tb_device_id,
    variable=variable, 
    start_ts=start_ts,
    end_ts=end_ts
)
```
And the same with any other Thingsboard related URL. A specific section about Thingsboard stuff in the settings would be the best, so it is easy to find and replace wherever needed.

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
...
+    headers = {"X-Authorization": f"Bearer {token}"}
+
+    response = requests.get(url, headers=headers)
+
+    logger.error(f"Request URL: {url}")
```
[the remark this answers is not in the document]
dalonsoa: This also looks like a very good plan, to be honest.

dalonsoa: Why are you logging here an error? There might be no errors.

Sahil590: This was just to see what was being passed was using for debugging

importing/utils.py
```diff
@@ -14,3 +23,55 @@ def remove_default_origins(apps, schema_editor):
     origins = ["file", "api", "Thingsboard"]
     for origin in origins:
         ImportOrigin.objects.filter(origin=origin).delete()
+
+
+def retrieve_thingsboard_device_id(token, customer_id, device_name):
+    """Retrieve ThingsBoard device ID for a given device name."""
+    api_headers = {"X-Authorization": f"Bearer {token}", "Accept": "application/json"}
+
+    devices_url = f"https://{settings.TB_HOST}/api/customer/{customer_id}/devices?pageSize=10000&page=0"
```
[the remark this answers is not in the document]
Sahil590: @dalonsoa would we want to specify the query limit ?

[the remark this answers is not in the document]
dalonsoa: To be honest, I don't know. I don't know how big the imports will be. Let's leave a limit, for now, and monitor if it becomes a problem when importing data from Thingsboard. I do not know if @ICHydro might have an opinion on this.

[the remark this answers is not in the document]
Sahil590: I can make it a setting and they can just change the ENV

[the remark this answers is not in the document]
Sahil590: will leave the default to 10,000

importing/utils.py
```diff
@@ -25,12 +24,17 @@ def remove_default_origins(apps, schema_editor):
...
     """Retrieve ThingsBoard device ID for a given device name."""
     api_headers = {"X-Authorization": f"Bearer {token}", "Accept": "application/json"}
 
-    devices_url = f"https://{settings.TB_HOST}/api/customer/{customer_id}/devices?pageSize=10000&page=0"
-    print(f"Using customer endpoint (customer ID: {customer_id})")
+    if settings.TB_CUSTOMER_DEVICES_URL is None:
+        raise Exception("TB_CUSTOMER_DEVICES_URL environment variable is not set.")
+    else:
+        devices_url = settings.TB_CUSTOMER_DEVICES_URL.format(customer_id=customer_id)
+    logger.debug(f"Using customer endpoint (customer ID: {customer_id})")
```
dalonsoa, line 35: As this is just to keep mypy happy - the URL is set in the settings, even if mypy does not know it - you can just add an assertion:

```python
assert settings.TB_CUSTOMER_DEVICES_URL is not None
devices_url = settings.TB_CUSTOMER_DEVICES_URL.format(customer_id=customer_id)
```

importing/utils.py
```diff
@@ -45,29 +49,48 @@ def retrieve_thingsboard_device_id(token, customer_id, device_name):
...
-        f"https://{settings.TB_HOST}/api/plugins/telemetry/DEVICE/{tb_device_id}/values/timeseries"
-        f"?interval=60000&limit=10000&agg=NONE"
-        f"&keys={variable}&startTs={start_ts}&endTs={end_ts}"
+    logger.debug(
+        f"Retrieving ThingsBoard data for device {tb_device_id}, variable {variable}"
     )
+    if settings.TB_TIMESERIES_URL is None:
+        raise Exception("TB_TIMESERIES_URL environment variable is not set.")
+    else:
+        url = settings.TB_TIMESERIES_URL.format(tb_device_id=tb_device_id)
```
dalonsoa: As above, you do not really need this. Just assert that the value is not None and format the URL.

importing/views.py
```diff
@@ -394,10 +401,11 @@ def form_valid(self, form):
                 end_ts=int(end_date.timestamp() * 1000),
             )
 
-            filename = (
-                f"thingsboard_{tb_device_name}_{tb_variable}_"
-                f"{datetime.now().strftime('%Y%m%d%H%M%S')}.json"
-            )
+            safe_device = slugify(tb_device_name)[:50]
+            safe_variable = slugify(tb_variable)[:50]
+            timestamp = timezone.now().strftime("%Y%m%d%H%M%S")
```
dalonsoa, line 406: Why do you need to slugify this? And why only the first 50 elements?

Sahil590: It was a copilot suggestion above to avoid having long file names.

Issue #579, opened 2026-03-04
Implement comparing traces


dalonsoa:
In principle, only to compare two traces, but if it is easy enough to implement, multiple comparisons would also be OK. 

![Image](https://github.com/user-attachments/assets/d47924d6-4b9c-499e-9c44-4fd55c0f2772)

@ICHydro In terms of styling and content, when comparing multiple traces, having them plotted as dots, and moreover including value, max and min, might result in a way too crowded plot. I'd suggest that when doing comparisons, we just plot the value as lines (no dots or additional traces). What do you think?

dalonsoa:
Hi @ICHydro , any comment on this before we dive into the coding?

ICHydro:
Sorry for the delau, but yes that looks good to me!

Wouter

PR #591, opened 2026-03-30
Use band plots in reports

dalonsoa:
Replaces the scatter plot in the reports page by line plots with shaded areas indicating the maximum and minimum regions. This is a preliminary step to implement multiple traces plotting - see #579 - since using scatter plots in this case (each with their maximum and minimum values) will result in a very cluttered plot.

A couple of bug fixes related to a change in the API have also been made.

Now, it will look like this:

image


measurement/dash_apps/plots.py
```diff
@@ -161,3 +182,44 @@ def create_report_plot(
...
+    for idx in reversed(gap_indices):
+        nan_rows.append(
+            {
+                "time": data.loc[idx, "time"] - median_diff / 2,
+                "value": float("nan"),
+                "maximum": data.loc[idx, "maximum"],
+                "minimum": data.loc[idx, "minimum"],
+            }
+        )
+    data.drop(columns=["time_diff"])
```
Sahil590: Won't this still keep the column in the frame below?

tests/measurement/test_dash_apps.py
Sahil590, line 1: I noticed that the tests won't run this in CI since it's not the same Django tests like the rest.

Sahil590 (approved):
I left some comments below I think overall it makes sense and is working as expected.

measurement/dash_apps/plots.py
dalonsoa: Yep. Forgot to assign it to `data` again. Sorted, now.

tests/measurement/test_dash_apps.py
dalonsoa: Sorted, as well. It seems VSCode was happy with the missing `TestCase` part. Also sorted, now and checked that it works.

PR #592, opened 2026-04-14
add new sidebar to be able to select stations

### LINKED ISSUE #588: Select stations to view
The landing page should be refactored such that visible stations can be selected from a list on the left. Stations own by the user and public stations owned by others should be available in two blocks, displaying only a finite number of elements - eg 6 or 8 - and allowing scrolling to see more (so if there are 200 stations, they are not displayed in a really long list). There should be an option in each of the list to select "All" or "None".

Sahil590:
I have added a new sidebar to the landing page. You can now filter on the map which station(s) you would like to display on the map. If you are logged out you can only see the public stations. When logged in you can see the private and public stations as well as the stations the user logged in owns.

https://github.com/user-attachments/assets/d2761190-c5c3-435e-8e0a-adbcc30c5043

Close #588


djangomain/dash_apps/stations_map.py
```diff
@@ -1,44 +1,371 @@
...
+        list[dict[str, str]]: Option dicts with ``"label"`` and ``"value"``
+            keys, ready for use in ``dcc.Checklist``.
+    """
+    options = []
+    for code in sorted(codes):
+        try:
+            obj = Station.objects.get(station_code=code)
+            label = f"{code} \u2013 {obj.station_name}" if obj.station_name else code
+        except Station.DoesNotExist:
+            label = code
```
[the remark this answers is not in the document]
Sahil590: I tried this but did not work. I can't really see any performance issues with the page loading.

djangomain/dash_apps/stations_map.py
```diff
@@ -1,44 +1,371 @@
...
+        block_id (str): Prefix used for internal control ids.
+        title (str): Text rendered in the card header.
+
+    Returns:
+        dbc.Card: Card containing an All/None button group and a scrollable
+            checklist.
+    """
+    return dbc.Card(
+        [
+            dbc.CardHeader(title, className="py-2"),
```
dalonsoa, line 95: What are `clasName`s for? Out of curiosity, mainly.

djangomain/dash_apps/stations_map.py
```diff
@@ -1,44 +1,371 @@
...
+
+    Returns:
+        list[dict[str, str]]: Option dicts with ``"label"`` and ``"value"``
+            keys, ready for use in ``dcc.Checklist``.
+    """
+    options = []
+    for code in sorted(codes):
+        try:
+            obj = Station.objects.get(station_code=code)
+            label = f"{code} \u2013 {obj.station_name}" if obj.station_name else code
```
dalonsoa, line 173: Why do you need `\u2013` instead of just a `-`?

djangomain/dash_apps/stations_map.py
```diff
@@ -1,44 +1,371 @@
...
+        except Station.DoesNotExist:
```
dalonsoa, line 174: If the station does not exist, then it should not appear as an option at all. Use [`supress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) to ignore the exception.

Not that the situation should ever arise, as I assume that the input codes are coming already from a query to the database, so all correspond to stations that do exist.

djangomain/dash_apps/stations_map.py
```diff
@@ -1,44 +1,371 @@
...
+@app.callback(
+    Output("owned-checklist", "value"),
+    [
+        Input("owned-checklist", "options"),
+        Input("owned-select-all", "n_clicks"),
+        Input("owned-select-none", "n_clicks"),
+    ],
+    State("owned-btn-store", "data"),
+)
+def owned_selection(options, n_all, n_none, store):
```
dalonsoa, line 243: `n_all` is not used in this function - neither in the one for the public stations - so why is it an input? If it needs to be an input for some reason, Copilot comment makes sense and should be actioned. Actually, how does the `all` button works? There's no logic to select all stations anywhere, as far as I can tell.

djangomain/dash_apps/stations_map.py
```diff
@@ -1,44 +1,371 @@
...
+@app.callback(
+    Output("public-checklist", "value"),
+    [
+        Input("public-checklist", "options"),
+        Input("public-select-all", "n_clicks"),
+        Input("public-select-none", "n_clicks"),
+    ],
+    State("public-btn-store", "data"),
+)
+def public_selection(options, n_all, n_none, store):
```
dalonsoa, line 296: Same comment as above for `n_all` not being used.

dalonsoa (changes requested):
I've a few changes and questions, but this looks really good! I've learnt a lot about Dash in 30 by going through your code!

stephwills (commented):
This looks really cool! I don't have too much to add -- I found it a bit difficult to understand the logic behind how the All/None buttons work too, but it seems to all work when testing.

One minor comment -- I notice that when you select/deselect a station in the checklist, it resets the zoom each time. It may be easier for users if it persists between clicks?

djangomain/dash_apps/stations_map.py
Sahil590: This one just adds padding. I used the inline suggestions when making these so they are not really necessary

djangomain/dash_apps/stations_map.py
Sahil590: I missed this It was a inline suggestion

djangomain/dash_apps/stations_map.py
Sahil590: I tried to remove this and the one below but when I do the "all" and "none" buttons stop working.

Sahil590: It's something to do with dash expecting there to be an input by default(all) even though it's not used in the function It's still being used on load

Sahil590:
@stephwills Thanks for the suggestion I managed to change it so that the map does not update every time a new station is sleeted. 
Before I was using the plotly express `px.scatter_mapbox` which would render a new map every time a new item was selected/deselected. Now we use plotly graph objects `go.Scaterrmapbox` to render the map which now allows us to change what is being plotted live on the graph. 
@dalonsoa I have also addressed some of the comments you made above.

Sahil590:
I have just added an improvement to the checklist selection it now uses the callback_context to get the button press

djangomain/dash_apps/stations_map.py
```diff
@@ -47,17 +311,43 @@ def update_map(stations) -> px.line:
...
+    df = pd.DataFrame(rows, columns=[*keys, "type"])
+    traces = []
+    for group, color in _COLOR_MAP.items():
+        sub = df[df["type"] == group]
+        traces.append(
+            go.Scattermapbox(
+                lat=sub["station_latitude"],
+                lon=sub["station_longitude"],
+                mode="markers",
+                marker={"color": color},
```
dalonsoa: Increase a bit the size of the marker, here:

```suggestion
                marker={"color": color, "size": 10},
```

dalonsoa (changes requested):
This looks good, and now I understand better the need for the inputs and how the buttons work. 

Now, I've two issues:
- The map extends way out of the bottom of the screen. Ideally, we want it to have a rectangular, horizontal shape and fit in a typical-sized screen. It might have been like that in the past, but this is a good opportunity to fix it.
- If you select no stations (all of them are deselected), the map disappear. This should not happen - the map should still be there, but without any station.

Sahil590:
I have addressed the changes in the previous comment

dalonsoa (changes requested):
There's still something not quite right, here. See screenshot below. 

If I am logged, I have both private and public stations (they are not really public, I have access because I am admin), and the map disappears if there are none selected.

Screenshot from 2026-04-24 07-07-09

We need the map to be there even when there's nothing selected. I do not know how to fix that, but before this PR, there was a map even with no stations, so it should be possible to do.

Sahil590:
@dalonsoa that's strange just tested it now on this branch, and it works for me and was able to try it on a another laptop.
image
@stephwills could you check if this works for you?

stephwills:
> @dalonsoa that's strange just tested it now on this branch, and it works for me and was able to try it on a another laptop. image @stephwills could you check if this works for you?

I'll check

stephwills:
> @dalonsoa that's strange just tested it now on this branch, and it works for me and was able to try it on a another laptop. image @stephwills could you check if this works for you?



> @dalonsoa that's strange just tested it now on this branch, and it works for me and was able to try it on a another laptop. image @stephwills could you check if this works for you?

Yes the map is still displaying for me if I have nothing selected

dalonsoa:
Then that's me. I must have done something wrong...

dalonsoa:
I'll re-check as soon as possible.

dalonsoa (approved):
I did not have my local branch up to date :(  Sorry about that. 

All looks good!

PR #593, opened 2026-04-15
Add plot resampling to validation app

### LINKED ISSUE #399: But, maybe, let's open an issue and implement that in a separate PR, once this approach is fully tested in the Report app.
Implement resampler in the validation app
_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/paricia/issues/362#issuecomment-2422071043_

stephwills:
This PR adds plot resampling to the validation app, similar to what is already implemented for the reporting app.

- I have followed the general implementation in the reporting app
- `plot_refresh_required` is additionally set to `True` if the plot is updated
  - `MAX_POINTS` is used to sample `plot_data`, also set to 1000 here 
- `lru_cache` is added to `generate_validation_report` to cache results

Closes #399

### DIFF
--- a/measurement/dash_apps/daily_validation.py
+++ b/measurement/dash_apps/daily_validation.py
@@ -6,10 +6,10 @@ import dash_bootstrap_components as dbc
 import pandas as pd
 from dash import Input, Output, State, dcc, html
 from dash_ag_grid import AgGrid
+from django.conf import settings
 from django_plotly_dash import DjangoDash
 from plotly import graph_objs as go
 
-from djangomain.settings.settings import MAX_POINTS
 from variable.models import Variable
 
 from ..filters import (
@@ -596,7 +596,7 @@ def callbacks(
                 plot_data = plot_data[
                     (plot_data["time"] >= start) & (plot_data["time"] <= end)
                 ]
-            every = max(1, len(plot_data) // MAX_POINTS)
+            every = max(1, len(plot_data) // settings.MAX_POINTS)
             plot_data = plot_data.iloc[::every]
             out_plot = create_validation_plot(
                 data=plot_data,
--- a/measurement/dash_apps/daily_validation.py
+++ b/measurement/dash_apps/daily_validation.py
@@ -36,6 +36,7 @@ app = DjangoDash(
 SELECTED_DAY: date | None = None
 DATA_SUMMARY: pd.DataFrame = pd.DataFrame()
 DATA_GRANULAR: pd.DataFrame = pd.DataFrame()
+MAX_POINTS = 1000
 
 # Filters
 filters_row1 = html.Div(
@@ -337,6 +338,7 @@ app.layout = html.Div(
         Input("save-button", "n_clicks"),
         Input("detail-date-picker", "date"),
         Input("plot_radio", "value"),
+        Input("plot", "relayoutData"),
     ],
     [
         State("tabs", "value"),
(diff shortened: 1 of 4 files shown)


measurement/dash_apps/daily_validation.py
```diff
@@ -36,6 +36,7 @@
 SELECTED_DAY: date | None = None
 DATA_SUMMARY: pd.DataFrame = pd.DataFrame()
 DATA_GRANULAR: pd.DataFrame = pd.DataFrame()
+MAX_POINTS = 1000
```
dalonsoa: Now that this is used in more than one place, I'd suggest we add it to the Django settings rather than "hiding" it within separate modules.

measurement/dash_apps/daily_validation.py
```diff
@@ -580,8 +588,17 @@ def callbacks(
...
+                start = in_plot_relayout_data["xaxis.range[0]"]
+                end = in_plot_relayout_data["xaxis.range[1]"]
+                plot_data = plot_data[
+                    (plot_data["time"] >= start) & (plot_data["time"] <= end)
+                ]
+            every = max(1, len(plot_data) // MAX_POINTS)
+            plot_data = plot_data.iloc[::every]
             out_plot = create_validation_plot(
-                data=DATA_GRANULAR,
+                data=plot_data,
```
dalonsoa: This is what is done in the reports page, but now that I see it, I think it can be done a bit more efficient. Copying a potentially huge dataframe is expensive, so let's minimize that. I'd suggest something like:


```suggestion
            npoints = len(DATA_GRANULAR)
            if "xaxis.range[0]" in in_plot_relayout_data:
                start = in_plot_relayout_data["xaxis.range[0]"]
                end = in_plot_relayout_data["xaxis.range[1]"]
                npoints = sum(
                    (DATA_GRANULAR["time"] >= start) & (DATA_GRANULAR["time"] <= end)
                )
            every = max(1, npoints // MAX_POINTS)
            out_plot = create_validation_plot(
                data=DATA_GRANULAR.iloc[::every],
```

I believe this should have the same functionality but be way more efficient, as we are working with views of the original dataframe, as well as a sum over boolean data.

dalonsoa (changes requested):
Looks good to me, but since we are on it, I believe we can do better that what I did before.

measurement/dash_apps/daily_validation.py
stephwills: Oh yes thanks, it shouldn't need copying so I removed that. 

I think missing this line 
```
plot_data = plot_data[(plot_data["time"] >= start) & (plot_data["time"] <= end)]
```
causes a problem as the df has to be sliced if the start and end dates have changed.

measurement/dash_apps/daily_validation.py
stephwills: Thanks -- made that change.

measurement/dash_apps/daily_validation.py
```diff
@@ -9,6 +9,7 @@
 from django_plotly_dash import DjangoDash
 from plotly import graph_objs as go
 
+from djangomain.settings.settings import MAX_POINTS
```
dalonsoa: The correct way of using Django settings is:

```python
```

and then use `settings.MAX_POINTS`. Check https://docs.djangoproject.com/en/6.0/topics/settings/#using-settings-in-python-code

measurement/dash_apps/daily_validation.py
dalonsoa: Very true.

measurement/dash_apps/daily_validation.py
stephwills: Ah ok thanks, I've changed that now

dalonsoa:
When you are done with making changes, please request re-review.

dalonsoa (approved):
LGTM!

PR #594, opened 2026-04-15
Hide edit and delete buttons for non-owner

### LINKED ISSUE #446: remove "edit" button if an object is not editable

The "edit" button appears at the top of a page of an object, even if the user is not the owner and therefore cannot edit the object. Clicking the edit button results in a "page not found" error.

The button should not appear if the user has no edit rights on the object.

stephwills:
This PR hides the edit and delete buttons in the object detail views if the user is not the owner or admin.

Closes #446

### DIFF
--- a/templates/object_detail.html
+++ b/templates/object_detail.html
@@ -11,8 +11,10 @@
   <div class="row">
     <div class="col">
       <a href="{% url list_url %}" class="btn btn-success">Return to list</a>
-      <a href="{% url edit_url pk %}" class="btn btn-warning">Edit</a>
-      <a href="{% url delete_url pk %}" class="btn btn-danger">Delete</a>
+      {% if can_edit or user.is_superuser %}
+        <a href="{% url edit_url pk %}" class="btn btn-warning">Edit</a>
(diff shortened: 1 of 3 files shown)


stephwills:
I am not currently sure why, but when I run the tests within `management.test_views` individually it causes errors, but they pass when the entire test suite is run.

templates/object_detail.html
```diff
@@ -11,8 +11,10 @@ <h2>
   <div class="row">
     <div class="col">
       <a href="{% url list_url %}" class="btn btn-success">Return to list</a>
-      <a href="{% url edit_url pk %}" class="btn btn-warning">Edit</a>
-      <a href="{% url delete_url pk %}" class="btn btn-danger">Delete</a>
+      {% if can_edit or user.is_superuser %}
```
dalonsoa: I would check supperuser status in the backend, so here you just check if `can_edit` is true. This way, all the logic about who can edit things is together, rather than in two steps.

dalonsoa (approved):
I've a minor change, but it looks good otherwise, so approving.

templates/object_detail.html
stephwills: Ok thanks I've made that change so will merge now

PR #595, opened 2026-04-16
Add asterisk to labels of required fields

### LINKED ISSUE #449: Show to the user which attributes of an object are required
It is not clear to the user what attributes of an object are required when creating a new object. For example, a station needs a timezone, but does not need a region for the object to be created.

Make clear which entries are required, for example with a red asterisk

stephwills:
This PR edits field labels in the create views to show add an asterisk if the field is required.

asterisk_create_view

Closes #449


dalonsoa (approved):
That was easy! LGTM!

PR #596, opened 2026-04-20
Implement reprocess as a RadioSelect button

### LINKED ISSUE #374: Implement re-process with a button, rather than a checkbox

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/paricia/issues/364#issuecomment-2391345114_

stephwills:
This PR updates the reprocess option in the custom data import views.

- In the create and update views, it is implemented as a `RadioSelect` button (selecting Yes or No)

reprocess_button

Closes #374

### DIFF
--- a/importing/views.py
+++ b/importing/views.py
@@ -52,16 +52,6 @@ class DataImportDetailView(CustomDetailView):
 
     model = DataImport
 
-    def get_form(self):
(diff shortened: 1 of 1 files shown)


stephwills:
Closing this PR, superseded by #602

PR #600, opened 2026-04-22
Add secondary trace widgets

### LINKED ISSUE #597: Add widgets to define secondary trace
These widgets should be hidden until the checkbox is ticked. Once made visible, they should appear populated with the same content than the main widgets (temporality, stations and variables). Obviously, they should also get invisible if the checkbox is un-ticked.

dalonsoa:
These widgets are shown only when the "Add secondary trace" button is toggled. The new widgets are populated at the same time as the primary ones, but kept hidden. The date range has not been duplicated as that is common and should correspond to the primary data time range, always.

Some refactoring has been made to reuse the widgets code. Also, a bit of styling has been added to the buttons, as they were a bit dry.

[Screencast from 2026-04-22 07-18-44.webm](https://github.com/user-attachments/assets/8c06b0f8-9f2a-404e-833d-b8a80225ab78)

Close #597

### DIFF
--- a/measurement/dash_apps/data_report.py
+++ b/measurement/dash_apps/data_report.py
@@ -19,45 +19,76 @@ from .plots import (
 MAX_POINTS = 1000
 """Maximum number of points to display in the graph."""
 
+
+def _traces_selection(block_id):
+    """Build the traces selection widgets.
+
+    Args:
+        block_id (str): Prefix used to build the component id.
+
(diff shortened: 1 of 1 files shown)


stephwills (approved):
All looks good to me! The widgets/buttons look good.

Sahil590 (approved):
LGTM

PR #601, opened 2026-04-22
Add second trace to plot.

### LINKED ISSUE #579: Implement comparing traces

### LINKED ISSUE #598: Plot second trace
A second axis should be added to the right hand side to display the secondary trace. There should be a second axis even if the magnitude to represent is the same, clearly identifying which is which. The secondary trace should be resampled as needed, like the primary trace.

dalonsoa:
Some of the code is refactored to avoid duplication when adding traces, but otherwise it tries to re-use as much of the existing functionality as possible. 

Some aspects of the layout have been tweaked to make the plot bigger and fill as much as of the view port as possible. There are other aspects related to the L&F and layout that could be improved, but that can be addressed as part of a separate issue. 

I only have one dataset to play with, so in the video below the second trace just overlaps with the first one, but otherwise it should work well with other traces.
 
[Screencast from 2026-04-22 11-57-43.webm](https://github.com/user-attachments/assets/9f9443df-5007-4e11-a8f7-e8b8b8b05e2f)

Close #598 
Close #579


measurement/dash_apps/plots.py
```diff
@@ -156,30 +169,31 @@ def create_report_plot(
                 fill="tonexty",
                 showlegend=False,
             ),
-        ]
+        ],
+        secondary_ys=[secondary_data, secondary_data, secondary_data],
     )
```
stephwills, line 174: I am slightly confused by the code here: `add_traces` is just adding one trace to the figure? What does the `secondary_ys` argument represent here?

stephwills: I guess it is because the Maximum and Minimum are being added as well?

measurement/dash_apps/data_report.py
```diff
@@ -255,19 +321,21 @@ def download_csv_report(
...
                 ),
+                True,
                 [],
             )
         except Exception as e:
             alert = dbc.Alert(f"Could not export data to CSV: {e}", color="warning")
-            return None, [alert]
+            return None, False, [alert]
 
     return None, []
```
stephwills: Should there be three things returned here? Like `None, [], True`

measurement/dash_apps/data_report.py
```diff
@@ -255,19 +321,21 @@ def download_csv_report(
                     content=file,
                     filename=f"{station}_{variable}_{temporality}_{start_time}-{end_time}.csv",
                 ),
+                True,
                 [],
```
stephwills, line 325: Should `True` and `[]` be the other way round? And on line 329

measurement/dash_apps/data_report.py
stephwills: I know this hasn't changed in this PR, but should this be `if figure["layout"]["title"]["text"] == "No data to plot":`

stephwills (approved):
All the secondary trace code looks good to me and seems to work quite nicely! I've left a few questions/comments.

They are more about some of the code to raise the alerts in `download_csv_report` and `update_alert`. I was trying to get the alerts to appear in the web app and not sure if there are problems with the existing code (or my understanding with when they are supposed to appear).

stephwills:
> All the secondary trace code looks good to me and seems to work quite nicely! I've left a few questions/comments.
> 
> They are more about some of the code to raise the alerts in `download_csv_report` and `update_alert`. I was trying to get the alerts to appear in the web app and not sure if there are problems with the existing code (or my understanding with when they are supposed to appear).

Apologies I had meant to ask these before approving and selected approval by accident

measurement/dash_apps/plots.py
dalonsoa: Exactly. Maximum and Minimum are needed to create the grey band around the data. Each of them is a trace in it self, and the space between them filled.

measurement/dash_apps/data_report.py
dalonsoa: Indeed. And the same in the other line. Most likely, that is why you are not getting the alerts.

measurement/dash_apps/data_report.py
dalonsoa: Yes!

measurement/dash_apps/data_report.py
dalonsoa: Indeed

stephwills (commented):
Sorry I am now having a problem seeing the data, including when the page is first loaded. I think the problem is that if there is "No data to plot" and the buttons are hidden, then I can't get the graph to update, even when a valid date is selected in the DatePicker. Does the same happen for you?

dalonsoa:
OK, I think it is sorted, now, but the whole "alerts" system is a bit hacky. Might be worth to have a dedicated issue about it.

dalonsoa:
@stephwills , @Sahil590 , are you both happy about how this one works, now?

stephwills:
> @stephwills , @Sahil590 , are you both happy about how this one works, now?

I am happy to approve. I think it would be good to open an issue about the alerts/buttons, as I think there may still be a problem if you select dates where there is no data, causing the buttons to disappear and then not being able to display data/use the buttons when switching back to valid dates. But I think that is separate to the secondary trace functionality here.

Sahil590 (approved):
Fine to approve. Had the same comment as Steph above but can be in a another issue.

PR #602, opened 2026-04-22
Remove reprocess field from data import and add as button in Edit view

### LINKED ISSUE #374: Implement re-process with a button, rather than a checkbox

stephwills:
This PR implements reprocess as a button in the `DataImportEditView` instead of as a checkbox.

After discussion in #374 I've implemented this as follows:
- I've removed the 'reprocess' field from the `DataImport` model (and so removed its use/mention elsewhere)
- I've added a button in the `DataImportEditView` which will trigger reprocessing (by setting the import status to "N"). I have only added it in the Edit view and not the Create view because I thought it makes more sense only there? I can change if needed.
- This is also triggered if the file is changed (and update instead of reprocess button used)
- The button is added alongside the other buttons at the top of the view

reprocess_button_2

I'm not completely sure I've implemented this in the best way so let me know if you would like it done differently. I had originally just changed the reprocess checkbox to a `RadioSelect` widget, which can be found in draft PR #596.

This would also require changes in the docs but I wanted to see if this is the right change first.

Closes #374


tests/importing/test_views.py
```diff
@@ -1127,3 +1101,74 @@ def test_missing_tb_variable(self):
...
+                "format": self.format.pk,
+                "rawfile": SimpleUploadedFile(
+                    "new_file.csv", b"01/01/2024 14:35:10,20.5\n"
+                ),
+                "observations": "",
+            },
+        )
+        self.assertEqual(response.status_code, 302)
+        self.data_import.refresh_from_db()
+        self.assertEqual(self.data_import.status, "N")
```
dalonsoa, line 1151: What's the default value for the status? Maybe check that it is indeed that one before updating the file, to make sure it has actually changed.

dalonsoa (approved):
I think this is very nice and much cleaner than the previous version. Please, update the docs and this is ready to go.

tests/importing/test_views.py
stephwills: Good point, the default value is "N" so I'll set it as "C" in `setUp` and check before/after the change

stephwills:
> I think this is very nice and much cleaner than the previous version. Please, update the docs and this is ready to go.

@dalonsoa did you want to check the changes to the docs?

docs/es/importing_data.md
```diff
@@ -46,7 +46,7 @@ Una vez que el formulario esté completo, haga clic en `Save` en la parte superi
 
 ![Ingestión de datos completada](assets/images/importing_completed.png)
 
-- **Error**: la ingesta de datos falló. La información sobre lo que salió mal debería aparecer en el cuadro de registro en la parte inferior del detalle de la importación de datos. Intente solucionar los problemas, según los comentarios proporcionados, marque la casilla "Reprocesar datos" y guarde el formulario nuevamente para activar otro proceso de ingesta de datos.
+- **Error**: la ingesta de datos falló. La información sobre lo que salió mal debería aparecer en el cuadro de registro en la parte inferior del detalle de la importación de datos. Para solucionar los problemas, según los comentarios proporcionados, marquen `Edit` y guarden el formulario nuevamente para actualizar. Para empezar otra ingesta de datos, marquen la casilla `Reprocess` en la parte de arriba de la pagina.
```
dalonsoa, line 49: You know Spanish!

dalonsoa (approved):
LGTM!

docs/es/importing_data.md
stephwills: I had some help haha

PR #604, opened 2026-04-23
590 layer page

### LINKED ISSUE #590: Add layer page
The "Add layer" button should open a view - maybe a pop-up? - to allow the user to select a map that they own or is public. It should be a list of maps, by name for now, and the user can select one or more. These, when clicking on "Accept" will be added to the list of maps in the main page. 

For now, as placeholder, just list the tiff image we have, which will be loaded directly from file rather than from the database.

Sahil590:
Fixes #590 

I have added a new page where we can see all the map layers imported and perform CRUD operations similar to the normal import process. It is a simple from asking for a name, description and the file itself. Waiting on the example tiff files before I can add some checks for it (can be tackled in a different issue). 

You can access the page by selecting the button in the import page or navigating to 
http://localhost:8000/importing/map-layers/


importing/views.py
```diff
@@ -443,3 +443,38 @@ def form_valid(self, form):
...
+
+    model = MapLayerImport
+    fields = ["name", "description", "file"]
+
+
+class MapLayerCreateView(CustomCreateView):
+    """View to create a map layer import."""
+
+    model = MapLayerImport
+    fields = ["name", "description", "file"]
```
dalonsoa: IN these two, you also need to add the `visibility` as a field that can be changed/selected.

importing/models.py
```diff
@@ -217,3 +217,30 @@ def clean(self) -> None:
 
     def get_absolute_url(self):
         return reverse("importing:thingsboardimportmap_detail", kwargs={"pk": self.pk})
+
+
+class MapLayerImport(PermissionsBase):
```
dalonsoa, line 208: This is OK for now, but I suspect that at some point we are going to need to add more fields here, like coordinate system, latitude and longitude range and possibly a few others. 

Let's leave it for now, for the MVP.

dalonsoa (changes requested):
I can create, edit and delete map-layers, but I cannot see the list, complaining about a permission issue (even though I'm admin). I believe this might be related to not setting the object level permissions properly with a `post_save` signal. This is done with all other models in Paricia, including `DataImport`. Check for example `formatting/signals/handlers.py` and do the same in `importing/signals/handlers.py` using the `@receiver(post_save, sender=MapLayerImport)` decorator in the `set_object_permissions` function.

Also the `Import data` button at the top needs to be transformed into a drop down and within there include an `Import timeseries` (what we do now) and a new `Import map` (what you are implementing).

Sahil590:
Should be working now

stephwills (approved):
This looks good to me

importing/models.py
dalonsoa: Open an issue about this one and schedule it when appropriate.

dalonsoa:
> Also the `Import data` button at the top needs to be transformed into a drop down and within there include an `Import timeseries` (what we do now) and a new `Import map` (what you are implementing).

Also open an issue about this one.

dalonsoa (approved):
LGTM and works. I need to regenerate the database, for some reason.

PR #606, opened 2026-04-27
Change to ThingsboardImportMap to inherit from PermissionsBase

### LINKED ISSUE #576: Thingsboardimportmap inherits from `Model` instead of `PermissionsBase` like all the other models. This means that it does not have the right permissions and ownership.
I've just noticed that this inherits from `Model` instead of `PermissionsBase` like all the other models. This means that it does not have the right permissions and ownership.

Probably it can be done in a separate PR, where we review that tings make sense, permission-wise (eg. people should not be able to use variables or stations they have not permission to use when creating the maps), so please open an issue about it to tackle in the future.

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/paricia/pull/570#discussion_r2863005771_

stephwills:
This PR changes `ThingsboardImportMap` to inherit from `PermissionsBase` instead of `models.Model`. Django needs a default to be provided for the `owner` field, to fill empty rows in the database. Is the following strategy ok (in the migrations file): set null=True initially for the owner field, fill in existing rows to a default value (currently the `station.owner`) and then change null back to False? I am guessing the Thingsboard code hasn't been deployed yet so this won't be used in the production database anyway?

Closes #576


importing/migrations/0015_thingsboardimportmap_owner_and_more.py
```diff
@@ -0,0 +1,44 @@
...
+def set_default_owner(apps, schema_editor):
+    ThingsboardImportMap = apps.get_model('importing', 'ThingsboardImportMap')
+
+    for tb_map in ThingsboardImportMap.objects.filter(owner__isnull=True).iterator():
+        tb_map.owner = tb_map.station.owner
+        tb_map.save()
+
+
+def reverse_set_default_owner(apps, schema_editor):
+    pass
```
dalonsoa, line 29: Add some docstrings to these functions to clarify their purpose, otherwise it is a bit unclear why you have done this.

dalonsoa (approved):
This is actually pretty neat. Well done! I didn't thought about this approach and might work quite well for other cases as well.

importing/migrations/0015_thingsboardimportmap_owner_and_more.py
stephwills: Ok thanks -- will add these and merge

Issue #599, opened 2026-04-22
[BUG] Tiles showing 403 in maps


dalonsoa:
A few weeks ago we observed the following not so nice landing page in Paricia, in the production(ish) server:

<img width="1325" height="692" alt="Image" src="https://github.com/user-attachments/assets/8a3d3748-eb73-40a5-bea3-98d64baea314" />

Some policy changes on the OSM side have taken place (see https://wiki.openstreetmap.org/wiki/Blocked_tiles and links therein). As we use this via Plotly/Dash, not directly, we might need to wait until something happens on that side, but it will be good to have a look at any application we develop that uses Open Street Map, to make sure they are still working as they should.

Having said that, there might be something we could do, so the issue requires further clarification. 

There was some discussion on the topic [here (private Teams channel)](https://teams.cloud.microsoft/l/message/19:f7518d9d650f49328f02a7fcaad4b140@thread.skype/1774851245953?tenantId=2b897507-ee8c-4575-830b-4f8267c3d307&groupId=2f67e4b4-18bf-4b93-9c4a-8cff86dad743&parentMessageId=1774851245953&teamName=RSE&channelName=Rubber%20Duck%20&createdTime=1774851245953).

stephwills:
There seems to be work towards a fix in this plotly [issue](https://github.com/plotly/plotly.py/issues/5562)

dalonsoa:
Great! I guess we just need to wait...

dalonsoa:
Seems to be solved upstream (in OSM).

PR #607, opened 2026-04-28
add layers to map

### LINKED ISSUE #589: Display spacial layers
A third block should be added to handle spatial data. As a default, the OSM should be there, selected, but a couple of buttons should allow a user to add, hide and remove layers to the map. These layers should always be display below the stations points.

Sahil590:
In this PR. 
I have added more functionality to the map page. 
Firstly, I added a way to select the different layer could be useful for this issue #599 
I have also added another block to the sidebar where you can select layers that have been created in the `MapLayerImport` from a dropdown, and it will display the TIF file that was uploaded to that entry. 
To make this possible we first need to read the data from the TIF file using the package `tifffile` and gets the info Then the coordinates are calculated in the `_compute_geotiff_coordinates`. 

The next few bits for rendering the image on to the map I used copilot for. I couldn't find much on rendering a tif file as an image for dash specifically so it helped with that. It also helped with adding the colour to the image. I am still exploring more options and found rasterio seems to work but still figuring it out.


djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +244,315 @@ def _station_block(block_id, title):
...
+    try:
+        with tifffile.TiffFile(file_path) as tif:
+            if not tif.pages:
+                raise ValueError("TIFF file has no pages.")
+
+            first_page = tif.pages[0]
+            coordinates = _extract_geotiff_coordinates_from_tifffile_page(first_page)
+            image_data = _array_to_png_data_uri(first_page.asarray())
+    except Exception as exc:
+        raise ValueError("Layer file is not a valid georeferenced GeoTIFF.") from exc
```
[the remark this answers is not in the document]
Sahil590: I can make this another issue once I get an answer from ic_hydro and have the checks when uploading the file in the form

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +244,315 @@ def _station_block(block_id, title):
...
+        [left, top],
+        [right, top],
+        [right, bottom],
+        [left, bottom],
+    ]
+    if not all(_is_valid_lon_lat(lon, lat) for lon, lat in coordinates):
+        raise ValueError("GeoTIFF coordinates must be valid lon/lat values.")
+
+    return [[float(lon), float(lat)] for lon, lat in coordinates]
+
```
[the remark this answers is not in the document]
Sahil590: still need to add tests for this awaiting some comments

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +244,315 @@ def _station_block(block_id, title):
...
+        return {}
+
+    try:
+        queryset = get_objects_for_user(
+            user,
+            "importing.view_maplayerimport",
+            klass=MapLayerImport,
+        )
+    except Exception:
+        return {}
```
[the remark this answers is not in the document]
Sahil590: fixed

djangomain/dash_apps/stations_map.py
```diff
@@ -1,23 +1,50 @@
...
+    "station_name",
+    "station_latitude",
+    "station_longitude",
+)
+_DEFAULT_MAP_STYLE = "carto-positron"
+_MAP_STYLE_OPTIONS = [
+    {"label": "Carto Positron", "value": "carto-positron"},
+    {"label": "Carto Darkmatter", "value": "carto-darkmatter"},
+    {"label": "OpenStreetMap", "value": "open-street-map"},
+]
```
dalonsoa, line 36: This is nice. I only find a bit not user friendly that when selecting a new map style, the existing layer disapears and removing and adding it again does not show it. The only way this works is if the map style is changed first in a fresh page, and then the layer is added.

djangomain/dash_apps/stations_map.py
```diff
@@ -82,16 +109,7 @@ def _all_none_buttons(block_id):
...
-
-    Args:
-        block_id (str): Prefix used for internal control ids.
-        title (str): Text rendered in the card header.
-
-    Returns:
-        dbc.Card: Card containing an All/None button group and a scrollable
-            checklist.
-    """
+    """Build a station-selection card containing controls and checklist."""
```
dalonsoa: Even though we have not been consistent with this in Paricia, we should try to keep adding type annotations for arguments and outputs, as well as their description in the docstrings.

djangomain/dash_apps/stations_map.py
```diff
@@ -104,11 +122,85 @@ def _station_block(block_id, title):
     )
 
 
+def _spatial_data_block():
+    """Build controls for selecting and managing spatial layers."""
+    return dbc.Card(
+        [
+            dbc.CardHeader("Spatial Data", className="py-2"),
+            dbc.CardBody(
```
dalonsoa, line 133: With these, you can select layers from a drop down and add them, but there is no way of seeing which layers are already present. I found the tickbox - much like how stations work - more intuitive to use than dropdown.

I'd suggest you keep the dropdown to select existing layers you can add, but those added appear in a list with checkboxes, like the stations, hidding them by unchecking the box.

djangomain/dash_apps/stations_map.py
```diff
@@ -104,11 +122,85 @@ def _station_block(block_id, title):
...
+                                color="secondary",
+                                outline=True,
+                            ),
+                            dbc.Button(
+                                "Hide",
+                                id="spatial-layer-hide",
+                                size="sm",
+                                color="secondary",
+                                outline=True,
+                            ),
```
dalonsoa: If you use checkboxes, as suggested above, this is not needed.

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +245,340 @@ def _station_block(block_id, title):
...
-    """Normalise a callback input value to a plain Python list.
-
-    Args:
-        value: Input that may be ``None``, a scalar, or a list.
-
-    Returns:
-        list: Empty list for falsy input; the original list; or a single-item
-            list wrapping a scalar value.
-    """
+    """Normalise a callback input value to a plain Python list."""
```
dalonsoa: Same as above. We want type annotations in the function signature and detail description of the arguments and output in the docstring.

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +245,340 @@ def _station_block(block_id, title):
...
     if not value:
         return []
     return value if isinstance(value, list) else [value]
 
 
 def _build_options(codes):
-    """Build Dash checklist option dicts from an iterable of station codes.
+    """Build checklist options from station code values."""
```
dalonsoa: Same as above. Why are you removing all docstrings, by the way?

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +245,340 @@ def _station_block(block_id, title):
...
+    rows = []
+    for code in selected_codes:
+        row = station_rows.get(code)
+        if not row:
+            continue
+        rows.append({**row, "type": station_group})
+    return rows
+
+
+def _normalise_spatial_layers(layers_raw):
```
dalonsoa: All of these functions to read and manipulate tiff data should go in its own module. In here we should keep only the page structure and callbacks.

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +245,340 @@ def _station_block(block_id, title):
...
+
+    return _compute_geotiff_coordinates(
+        page.imagewidth,
+        page.imagelength,
+        pixel_scale_tag.value,
+        tiepoint_tag.value,
+    )
+
+
+def _scale_to_uint8(values, finite_mask):
```
dalonsoa: Why is this needed?

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +245,340 @@ def _station_block(block_id, title):
...
+    minimum = float(finite_values.min())
+    maximum = float(finite_values.max())
+    if maximum <= minimum:
+        return np.zeros(values.shape, dtype=np.uint8)
+
+    norm = (np.nan_to_num(values, nan=minimum) - minimum) / (maximum - minimum)
+    return np.clip(norm * 255, 0, 255).astype(np.uint8)
+
+
+def _apply_viridis_like(norm):
```
dalonsoa: Same here. It is OK to use Copilot to generate code, but what things do and why they are needed must be fully documented and have a purpose.

djangomain/dash_apps/stations_map.py
```diff
@@ -153,51 +245,340 @@ def _station_block(block_id, title):
...
+    stops = np.array(
+        [
+            [68, 1, 84],
+            [59, 82, 139],
+            [33, 145, 140],
+            [94, 201, 98],
+            [253, 231, 37],
+        ],
+        dtype=np.float32,
+    )
```
dalonsoa: Where are these arbitrary numbers coming from?

dalonsoa (changes requested):
OK, this seems to work, but there are quite a few things to change:

1. Many docstrings that were present are now gone. That is not good. They should be re-added, missing type annotations also added and, if needed, update the docstring if they were not correct.
2. All tiff-related manipulation functions should be in their own module.
3. All functions - specially those generated by Copilot - need to be fully documented, type annotated and have a reason to be. There is a lot of code at the moment whose purpose is unclear, and that cannot be.

I've also done a couple of funcitonality remarks in a couple of places.

dalonsoa:
Also, the tiff-related dependency should be added to the dev- and doc- requirements file, as well.

djangomain/dash_apps/stations_map.py
Sahil590: I noticed the same thing would still need to work on that

djangomain/dash_apps/stations_map.py
Sahil590: Yep was going to address the doc strings at the very end once i have the final functions

Sahil590:
@dalonsoa Thanks for the feedback on the PR I just got in touch with Ryan, and they did use the TIF files in the ESM, so I think it would be better if we used the same tool as them and will limit the AI generated code needed.

Sahil590:
@dalonsoa Rasterio only works with python 3.12+ can I upgrade the python version to 13 for the whole project?

dalonsoa:
Yes, please, do so. Upgrade to whatever is the newest version that works. Ideally 3.14, if nothing breaks.

Sahil590:
> Yes, please, do so. Upgrade to whatever is the newest version that works. Ideally 3.14, if nothing breaks.

Will open another PR for this

Sahil590:
I have made a lot more improvements to the code since last time. The original logic I found for building a PNG was quite complicated and wasn't really necessary. I found a library called [rio-tiler](https://github.com/cogeotiff/rio-tiler) this makes is much easier to manipulate the tif file and render a usable image on the map.

Sahil590:
@dalonsoa this should be ready now

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+            file_path = str(layer.file.path)
+        except (ValueError, OSError):
+            continue
+
+        layer_id = f"maplayer-{layer.pk}"
+        layer_index[layer_id] = {
+            "id": layer_id,
+            "name": str(layer.name),
+            "file_path": file_path,
+        }
```
dalonsoa: If the object exists, then it must have a file and an path. There's no other option, otherwise it should not have been a valid submission, so this check is overkilling.

```suggestion
        layer_id = f"maplayer-{layer.pk}"
        layer_index[layer_id] = {
            "id": layer_id,
            "name": str(layer.name),
            "file_path": str(layer.file.path),
        }
```

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+def available_map_layers_by_id(user):
+    """Return user-viewable GeoTIFF layers keyed by dash dropdown id.
+
+    Args:
+        user: Django user used to resolve object-level permissions.
+
+    Returns:
+        dict[str, dict[str, str]]: Layer metadata keyed by maplayer id. Returns
+            an empty dict when the user is None or lookup fails.
+    """
```
dalonsoa: Remember that we annotate the function signature (both, inputs and output), not the docstring. That needs to be changed all over the code.

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+    if str(src_crs) != _TARGET_MAPBOX_COORDS_CRS:
+        try:
+            left, bottom, right, top = transform_bounds(
+                src_crs,
+                _TARGET_MAPBOX_COORDS_CRS,
+                left,
+                bottom,
+                right,
+                top,
+                densify_pts=21,
```
dalonsoa: What does this mean?

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+    return [
+        [float(left), float(top)],
+        [float(right), float(top)],
+        [float(right), float(bottom)],
+        [float(left), float(bottom)],
+    ]
+
+
+def _build_image_payload(file_path):
+    """Read a georeferenced single-band GeoTIFF and render it for mapbox.
```
dalonsoa, line 101: What's a `georeferenced single-band GeoTIFF`?

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+        coordinates = _bounds_to_lonlat_coordinates(
+            img.bounds,
+            _TARGET_RASTER_RENDER_CRS,
+        )
+
+        min_value = float(img.array.min())
+        max_value = float(img.array.max())
+        if max_value > min_value:
+            img.rescale(in_range=((min_value, max_value),))
```
dalonsoa: What's the purpose of this? min is already the minimum value, and max the maximum, so what is to rescale? And what happens if max_value is smaller than min_value?

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+        bounds: Sequence of (left, bottom, right, top) in src_crs units.
+        src_crs: CRS the bounds are expressed in.
+
+    Returns:
+        list[list[float]]: Corner coordinates ordered [top-left, top-right,
+            bottom-right, bottom-left].
+    """
+    left, bottom, right, top = bounds
+
+    if str(src_crs) != _TARGET_MAPBOX_COORDS_CRS:
```
dalonsoa: You are calling this function with `_TARGET_RASTER_RENDER_CRS` so, by definition, `str(src_crs) != _TARGET_MAPBOX_COORDS_CRS` always. There is not really any check on the real coordinates of the image, as provided by the tiff file.

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+        dataset = src.dataset
+        if dataset is None:
+            raise ValueError("GeoTIFF dataset could not be opened.")
+        img = src.preview(
+            dst_crs=_TARGET_RASTER_RENDER_CRS,
+            max_size=4096,
+        )
+        coordinates = _bounds_to_lonlat_coordinates(
+            img.bounds,
+            _TARGET_RASTER_RENDER_CRS,
```
dalonsoa: Shouldn't you be passing `src.crs` to get the coordinate system the image is actually in?

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+            coordinates.
+    """
+    del mtime
+
+    try:
+        return _build_image_payload(file_path)
+    except ValueError:
+        raise
+    except (RioTilerError, Exception) as exc:
+        raise ValueError("Layer file is not a valid georeferenced GeoTIFF.") from exc
```
dalonsoa: If is it not a valid file, it should not be in the database, to start with. These checks should all be done when ingesting the data, not when using it. Open an issue to check the validity of the files submitted at the time of submission.

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+    map_layers = []
+    available_layers = available_map_layers_by_id(user)
+
+    for layer in layers_raw:
+        if not layer["visible"]:
+            continue
+
+        resolved_layer = available_layers.get(layer["id"])
+        if not resolved_layer:
+            continue
```
dalonsoa, line 184: How can this situation be possible?

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Args:
+        file_path: Absolute path to the GeoTIFF file on disk.
+        mtime: File modification time used as part of the cache key.
+
+    Returns:
+        dict[str, object]: Payload containing image data URI and map
+            coordinates.
+    """
+    del mtime
```
dalonsoa: Why do you delete it? If you are not using it and it is only for caching, just indicate so with an underscore in the function signature, `_mtime`.

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -0,0 +1,213 @@
...
+
+        resolved_layer = available_layers.get(layer["id"])
+        if not resolved_layer:
+            continue
+
+        try:
+            mtime = os.path.getmtime(resolved_layer["file_path"])
+            payload = load_geotiff_payload(resolved_layer["file_path"], mtime)
+        except (OSError, ValueError):
+            continue
```
dalonsoa, line 191: If there were an error, at the very least it should be logged, not just ignored. If you have your custom error logic in `load_geotiff_payload` and now you just ignore all that logic, what's the point of raising exceptions?

djangomain/dash_apps/stations_map.py
```diff
@@ -1,23 +1,38 @@
...
+from dash import ALL, MATCH, Input, Output, Patch, State, dcc, html, no_update
 from django_plotly_dash import DjangoDash
 
+from djangomain.dash_apps.geotiff_layers import (
+    available_map_layers_by_id,
+    build_mapbox_layers,
+)
 from station.models import Station
 
 SCROLL_HEIGHT = "280px"
```
dalonsoa: Reduce this a bit. At the moment, I feel is visually too large.

djangomain/dash_apps/stations_map.py
```diff
@@ -188,16 +265,142 @@ def _build_options(codes):
...
+
+    for item in _ensure_list(store_data):
+        if isinstance(item, dict):
+            layer_id = item.get("id")
+            visible = bool(item.get("visible", True))
+        else:
+            layer_id = item
+            visible = True
+
+        if layer_id in (None, "") or layer_id in seen:
```
dalonsoa, line 325: Why would layer id ever be None or empty?

dalonsoa: Or why would it be duplicated and already been seen?

djangomain/dash_apps/stations_map.py
```diff
@@ -248,7 +451,139 @@ def populate_options(all_raw, **kwargs):
...
+                _ensure_list(remove_ids),
+                _ensure_list(_remove_clicks),
+            )
+            if isinstance(button_id, dict) and clicks
+        ]
+        removed_id = (
+            max(clicked_remove_ids, key=lambda item: item[1])[0]
+            if clicked_remove_ids
+            else None
+        )
```
dalonsoa, line 542: What's this doing? Why is all this logic needed to get the removed id?

dalonsoa (changes requested):
The code does what it should, but I feel there's still a lot of unclear code, or taken "out of the box" without a clear purpose.

djangomain/dash_apps/geotiff_layers.py
Sahil590: Single band is when there is only one layer per pixel

djangomain/dash_apps/geotiff_layers.py
Sahil590: It comes from the warping https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html#rasterio.warp.transform_bounds

dalonsoa: The default of rasterio is already 21, so setting it explicitly does not add anything.

djangomain/dash_apps/geotiff_layers.py
Sahil590: It was using another error logger before forgot to amend it the logger defined

djangomain/dash_apps/geotiff_layers.py
Sahil590: Since we store the layers for the user in the browser they might have a layer that they have lost access to but can still enable due to the caching, so this prevents the case where the user is trying to load an unauthorised layer.

djangomain/dash_apps/geotiff_layers.py
Sahil590: Yeas I used the other before because we were using the preview but have changed it to pass the dataset from the file itself rather than the preview

djangomain/dash_apps/stations_map.py
Sahil590: Since we get callbacks from the client side it would be possible to tamper with the callback and send any value over.

djangomain/dash_apps/stations_map.py
Sahil590: The callback doesn't say which `X` was clicked to remove it only indicates if it has been clicked so we needed to have some logic to get which button was clicked most recently through the click count. Its not the best way of doing it so going to try explore a better way.

djangomain/dash_apps/geotiff_layers.py
Sahil590: Rescale is for converting the image into 8bit values that are needed for converting it into a PNG

Sahil590:
@dalonsoa I have made some more changes based on the comments above and replied to them. Although not sure why the docs are failing. Also the tests are not being run all the way.

djangomain/dash_apps/geotiff_layers.py
dalonsoa: As we have them in the function signature - where we want them - they should not be here. Otherwise we risk them getting offsync. The same applies everywhere else.

djangomain/dash_apps/geotiff_layers.py
```diff
@@ -124,17 +117,14 @@ def _build_image_payload(file_path):
...
+            dataset.bounds,
+            src.crs,
         )
 
-        min_value = float(img.array.min())
-        max_value = float(img.array.max())
-        if max_value > min_value:
-            img.rescale(in_range=((min_value, max_value),))
+        band_stats = next(iter(src.statistics().values()))
+        img.rescale(in_range=((band_stats.min, band_stats.max),))
```
dalonsoa, line 127: I'm still not entirely sure what this is about. Maybe add an inline comment for future generations to understand.

dalonsoa (commented):
Looks good, now. Address the docstring comment, add a few tests and I think this is ready to go!

djangomain/dash_apps/geotiff_layers.py
Sahil590: This would be the best description i could find https://github.com/developmentseed/titiler/discussions/494

djangomain/dash_apps/geotiff_layers.py
Sahil590: Docs still not passing the tests

Sahil590: image
This is the error we got. Should i go ahead and set the env `DISABLE_MKDOCS_2_WARNING=true` in the CI for this PR or open another?

dalonsoa: :( Yes please, add that env variable to the workflow and let's get rid of it. Also, if you don't mind, pin the mkdocs version to less than 2, with "mkdocs<2.0.0" in `pyproject.toml`. You will need to recreate the doc requirements file, as well.

dalonsoa: A separate PR, by the way. Sorry for the confusion.

djangomain/dash_apps/geotiff_layers.py
dalonsoa: Add a comment with a link to that discussion. Clearly, I'm not alone understanding what this is for.

Sahil590:
@dalonsoa added the remaining tests and the link to the discussion

dalonsoa (approved):
OK, this looks good to me! Let's wait for @stephwills in case she has any comments and we can merge next week. 

Well done! Lots of work!

stephwills (approved):
This looks really good to me and a lot of work! I don't have anything to add regarding implementation. Good to learn how the geotiff code works. 

As a side comment, would you be able to share the tiff file that you were using to test it out? Just so I can try it locally. I can't remember if this was shared elsewhere.

stephwills:
Is this ready to merge, or is the plan to wait for #621 to be fixed?

dalonsoa:
Merged!

PR #610, opened 2026-05-06
upgrade to python 3.14

### LINKED ISSUE #609: Upgrade Python
Yes, please, do so. Upgrade to whatever is the newest version that works. Ideally 3.14, if nothing breaks.

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/paricia/issues/607#issuecomment-4387740940_

Sahil590:
#609 
As mentioned in #607 we need >=3.12 to work with [rasterio](https://github.com/mapbox/rasterio/releases). I have tested all the functionality apart from the importing app and seems to be all working with 3.14.


dalonsoa:
It seems that something with the docs is not quite right. I'm going to be working on the documentation tomorrow, so I might be able to solve this myself.

Sahil590:
Yes seems to be with `lxml` package from `mkdocs-caption` I pinned the version that works in the toml file

pyproject.toml
```diff
@@ -80,7 +81,7 @@ explicit_package_bases = true
 
 [tool.ruff]
 exclude = ["*/migrations"]
-target-version = "py311"
+target-version = "py313"
```
dalonsoa: I believe this should be `py314`, right?

```suggestion
target-version = "py314"
```

dalonsoa (approved):
All looks good! Thanks for fixing the dependency issue for the docs.

pyproject.toml
Sahil590: I added this but getting this warning should i still go ahead with it?
image

Sahil590:
I noticed this when running the tests. Might be related to why this is failing #577.  

image

pyproject.toml
dalonsoa: Mmmm... merge, anyway, but don't remove the warning.

dalonsoa:
Yep. That has been in my todo list for a while. Could you open an issue to fix it?

PR #611, opened 2026-05-07
Batch bulk insert into db

### LINKED ISSUE #477: Analyse (and if possible, reduce) memory consumption of imports
The data import process consumes a lot of memory. When importing files of around 1 million records, memory consumption can exceed 8GB.

Understand what exactly consumes so much memory, and try to find ways to reduce it.

stephwills:
I believe a lot of the memory problem occurs when data is saved into the database rather than during the prior data processing (with Pandas). 

Currently, for each variable dataframe, the rows are converted into a list of dictionaries, then into model instances, and then bulk inserted into the database. This PR instead batches this process and avoids creating a full list of dictionaries. The batch size can be set in `settings.py`.

I have looked a little into using django-import-export but am not sure if it provides the flexibility needed given how the columns are mapped -- will look into it more but opening this PR for now.

Closes #477


Sahil590:
@stephwills there is this package https://django-import-export.readthedocs.io/en/latest/ we used it for [direct framework](https://github.com/direct-framework/direct-webapp/pull/613). Might be easier to bulk import into the db

stephwills:
> @stephwills there is this package https://django-import-export.readthedocs.io/en/latest/ we used it for [direct framework](https://github.com/direct-framework/direct-webapp/pull/613). Might be easier to bulk import into the db

Ah thanks, I'll take a look!

dalonsoa (approved):
This looks good, and if it works, fantastic. I wonder if the main advantage is not really the batches but having used `itertuples` and if this would have work equally nice just bye replacing the model instance creation by:

```python
model_instances = [create_and_clean(**record.as_dict()) for record in table.itertuples(index=False)]
```

stephwills:
> This looks good, and if it works, fantastic. I wonder if the main advantage is not really the batches but having used `itertuples` and if this would have work equally nice just bye replacing the model instance creation by:
> 
> ```python
> model_instances = [create_and_clean(**record.as_dict()) for record in table.itertuples(index=False)]
> ```

Oh yes good point, I will try it

dalonsoa:
> > This looks good, and if it works, fantastic. I wonder if the main advantage is not really the batches but having used `itertuples` and if this would have work equally nice just bye replacing the model instance creation by:
> > ```python
> > model_instances = [create_and_clean(**record.as_dict()) for record in table.itertuples(index=False)]
> > ```
> 
> Oh yes good point, I will try it

It might be that the bulk creation even works with an iterator (parenthesis instead of  brackets), which will save more memory. No idea. Try and see if it happens.

```python
model_instances = (create_and_clean(**record.as_dict()) for record in table.itertuples(index=False))
```

stephwills:
> > > This looks good, and if it works, fantastic. I wonder if the main advantage is not really the batches but having used `itertuples` and if this would have work equally nice just bye replacing the model instance creation by:
> > > ```python
> > > model_instances = [create_and_clean(**record.as_dict()) for record in table.itertuples(index=False)]
> > > ```
> > 
> > 
> > Oh yes good point, I will try it
> 
> It might be that the bulk creation even works with an iterator (parenthesis instead of brackets), which will save more memory. No idea. Try and see if it happens.
> 
> ```python
> model_instances = (create_and_clean(**record.as_dict()) for record in table.itertuples(index=False))
> ```

When I tried this^ the generator was better than the original code but consumes more memory than using the batches (I think because it casts the objects to a list still internally). You can set a `batch_size` argument within `bulk_create`, but it seems to be worse when I try that (not sure why..).

dalonsoa:
Interesting. Then, the approach is clear - let's stick with the batches!

PR #615, opened 2026-05-11
Fix update of secondary variables

### LINKED ISSUE #614: [BUG] No new variables in additional trace
**Describe the bug**
When adding an additional trace to the reports page, the variables available to choose from in the additional trace dropdown are those of the first station, not of the additional station. 

**Expected behavior**
The variables available should correspond to the second station, not the first one. They should be independent from one another. 

**Screenshots**

In the following screenshot, options of the secondary trace should be `Caudal` and `Nivel de agua`, not `Temperatura ambiente`, which is a variable of the primary station only.

Image

dalonsoa:
So we can select them independently on the first station/variable combination.

Fixes #614

### DIFF
--- a/measurement/dash_apps/data_report.py
+++ b/measurement/dash_apps/data_report.py
@@ -407,13 +407,16 @@ def populate_stations_dropdown(
         Output("secondary_variable_drop", "value"),
     ],
     Input("primary_station_drop", "value"),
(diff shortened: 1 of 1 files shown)


stephwills (approved):
Looks good to me!

PR #616, opened 2026-05-12
Update alerts

### LINKED ISSUE #605: Review alerts in the Reports page
> @stephwills , @Sahil590 , are you both happy about how this one works, now?

I am happy to approve. I think it would be good to open an issue about the alerts/buttons, as I think there may still be a problem if you select dates where there is no data, causing the buttons to disappear and then not being able to display data/use the buttons when switching back to valid dates. But I think that is separate to the secondary trace functionality here.

_Originally posted by @stephwills in https://github.com/ImperialCollegeLondon/paricia/issues/601#issuecomment-4326134053_

dalonsoa:
The alerts now are shown when clicking the buttons, not when changing the fields. Including specific warning information about why the data cannot be shown. The viewport size is changed a little bit such that the plot can be seen entirely during an alert.

Close #605

### DIFF
--- a/measurement/dash_apps/data_report.py
+++ b/measurement/dash_apps/data_report.py
@@ -105,7 +105,7 @@ buttons_div = html.Div(
         ),
     ],
     id="buttons_div",
-    hidden=True,
+    hidden=False,
     style={"margin-top": "30px"},
 )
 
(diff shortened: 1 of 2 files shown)


Sahil590 (approved):
LGTM.

stephwills (approved):
Looks good to me!

PR #617, opened 2026-05-14
Refactor importing functions

### LINKED ISSUE #198: Refactor `importing` functions
They have changed just to use the new structure of models, but they need a rigorous review and rewriting to make them efficient and follow best coding practices. 

https://github.com/ImperialCollegeLondon/paricia/blob/develop/importing/functions.py

(the whole `importing` app deserves a review, but that's a topic for another issue)

### LINKED ISSUE #612: Pandas upgrade to 3.0
I noticed this when running the tests. Might be related to why this is failing #577.  

image

_Originally posted by @Sahil590 in https://github.com/ImperialCollegeLondon/paricia/issues/610#issuecomment-4395291107_

stephwills:
This PR organises the functions in `importing/functions.py` (chunking them) and adds tests.

A summary of some of the main changes:

- Removed `validate_dates` and `get_last_uploaded_date` as they weren't being used 
- `read_file_excel` now working
- `process_datetime_column` and `standardise_datetime` - have changed to use pandas operations instead of applying the `standardise_datetime` function to each row
- `construct_matrix` - have split this up into separate functions; calls `get_processed_variable_data`, which does the processing for each variable type (in turn calling different processing functions)
- `standardise_floats` - have tried to simplify this again to use pandas operations instead of applying function to each row
- Upgraded `pandas` to `3.0.1` and added `openpyxl` to help with testing the `read_file_excel function` (removed `django-excel-response` as it wasn't being used and caused a conflict)

Let me know if anything doesn't make sense, particularly when switching functions to use pandas operations.

Closes #198 
Closes #612


importing/functions.py
```diff
@@ -180,28 +106,25 @@ def process_datetime_columns(
...
-        )
-        data.drop(columns=["datetime_str"], inplace=True)
+        data["date"] = data.iloc[:, cols].astype(str).agg(" ".join, axis=1)
+    else:  # If single column, rename to 'date'
+        data = data.rename(columns={data.columns[file_format.date_column]: "date"})
+
+    # Invalid dates become NaT
+    data["date"] = pd.to_datetime(
+        data["date"], format=dt_format, errors="coerce"
+    ).dt.tz_localize(tz)
```
stephwills, line 125: Ok to use `coerce` here? Should error be raised instead?

importing/functions.py
```diff
@@ -374,151 +265,210 @@ def construct_matrix(
...
+        The dataframe with data now standardised.
+    """
+    for col in data:
+        if col == "date":
+            continue
+        if classification.decimal_comma:
+            data[col] = pd.to_numeric(
+                data[col].astype(str).str.replace(".", "").str.replace(",", "."),
+                errors="coerce",
+            )
```
stephwills: Ok using `coerce` here?

importing/functions.py
dalonsoa: I think we should raise and error. If the date/time format selected is not the right one for the data being ingested, we do not want to keep injesting the data, but raise an error and let the user know what went wrong.

importing/functions.py
```diff
@@ -224,76 +147,38 @@ def read_data_to_import(source_file: Any, file_format: Format, timezone: str):
...
-    Returns:
-        A datetime object or None if date_time is not in a recognised format.
-    """
-    if isinstance(date_time, datetime):
-        return date_time
-    elif isinstance(date_time, np.datetime64):
-        date_time = datetime.fromtimestamp(
-            float((date_time - unix_epoch) / one_second), tz=zoneinfo.ZoneInfo("UTC")
-        )
-        return date_time
```
dalonsoa, line 244: Is the new approach able to handle timestamps?

importing/functions.py
```diff
@@ -374,151 +265,210 @@ def construct_matrix(
...
+        classification: a formatting.Classification object.
+
+    Returns:
+        A tuple of the validated DataFrame and a list of mappings for the columns that
+            have been validated, to be used in renaming.
+    """
+    columns = [("date", "date")]
+
+    # Validation of values; non-validated values are set to np.nan
+    columns.append((classification.value, "value"))
```
dalonsoa: I guess you can start with this:


```suggestion
    # Validation of values; non-validated values are set to np.nan
    columns = [("date", "date"), (classification.value, "value")]
```

importing/functions.py
```diff
@@ -374,151 +265,210 @@ def construct_matrix(
...
-    try:
-        val_str = val_str.replace(".", "")
-        val_str = val_str.replace(",", ".")
-        val_num = float(val_str)
-    except ValueError:
-        val_num = None
-    return val_num
+    data["value"] = data["value"].diff()
+    data.loc[data["value"] < 0, "value"] = np.nan
+    return data.dropna()
```
dalonsoa, line 366: This is not using the `classification_resolution` (which frankly, I've no idea what it is), that was potentially being applied in the previous version.

importing/functions.py
```diff
@@ -374,151 +265,210 @@ def construct_matrix(
...
+    data = remove_nan_rows(data, classification, columns)
+
+    if classification.incremental:
+        data = process_incremental_data(data)
+
+    if acc := classification.accumulate:
+        data = process_cumulative_data(data, classification, acc, start_date, end_date)
+    else:
+        if classification.resolution:
+            data["value"] = data["value"] * float(classification.resolution)
```
dalonsoa, line 444: Ah, OK, you handle it here.

importing/functions.py
dalonsoa: Same here. We should let user know if the chosen classification does not work for their data.

importing/functions.py
```diff
@@ -374,151 +265,210 @@ def construct_matrix(
...
+        if col == "date":
+            continue
+        if classification.decimal_comma:
+            data[col] = pd.to_numeric(
+                data[col].astype(str).str.replace(".", "").str.replace(",", "."),
+                errors="coerce",
+            )
+        else:  # if float commas
+            data[col] = pd.to_numeric(
+                data[col].astype(str).str.replace(",", ""), errors="coerce"
```
dalonsoa: And here.

tests/importing/test_functions.py
```diff
@@ -288,3 +552,119 @@ def test_process_datetime_columns_different_columns(self):
...
+        # Call the function
+        processed_data = process_datetime_columns(data, file_format, timezone)
+
+        # Assert the result
+        expected_data = pd.DataFrame(
+            {
+                "date": [
+                    pd.Timestamp("2022-01-01 09:00:00", tz="America/New_York"),
+                    pd.Timestamp("2022-01-01 10:00:00", tz="America/New_York"),
+                    pd.Timestamp("2022-01-01 11:00:00", tz="America/New_York"),
```
dalonsoa, line 594: It seems it does work with timestamps!

dalonsoa (changes requested):
This looks really, really much better. I've made a couple of comments, but otherwise, my only comment is about the use of `coerce`. It think that we should raise and error and inform the user of the problem, with as much detail as possible so they can fix it.

importing/functions.py
stephwills: Ok that makes sense!

importing/functions.py
stephwills: Ah yes it happens within `process_cumulative_data` for the cumulative data, and in `get_processed_variable_data` for the non-cumulative

dalonsoa (approved):
Perfect!

PR #620, opened 2026-05-21
pin mkdocs to less than 2.0 and add env to supress warning

### LINKED ISSUE #619: Pin mkdocs and supress warning
:( Yes please, add that env variable to the workflow and let's get rid of it. Also, if you don't mind, pin the mkdocs version to less than 2, with "mkdocs<2.0.0" in `pyproject.toml`. You will need to recreate the doc requirements file, as well.

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/paricia/pull/607#discussion_r3279739305_

Sahil590:
Fixes #619

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: postgres
+          DISABLE_MKDOCS_2_WARNING: true
         ports:
           - 5432:5432
         # needed because the postgres container does not provide a healthcheck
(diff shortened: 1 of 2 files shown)


dalonsoa (commented):
I'm not entirely sure, but you might need to re-generate `[requirements-doc.txt](https://github.com/ImperialCollegeLondon/paricia/blob/main/requirements-doc.txt)` to capture that we are pinning the mkdocs version?

Sahil590:
@dalonsoa I did run pip-compile but did not change anything its on the most up to date before 2.0

dalonsoa (changes requested):
All good, then!

stephwills (approved):
LGTM!

dalonsoa (approved):
Now

Issue #618, opened 2026-05-19
If is it not a valid file, it should not be in the database, to start with.


Sahil590:
If is it not a valid file, it should not be in the database, to start with. These checks should all be done when ingesting the data, not when using it. Open an issue to check the validity of the files submitted at the time of submission.

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/paricia/pull/607#discussion_r3264054579_

dalonsoa:
Fixed by https://github.com/ImperialCollegeLondon/paricia/pull/622

PR #622, opened 2026-05-21
608 file validation

### LINKED ISSUE #608: Add validation for file limits
Loading GeoTIFF data via `first_page.asarray()` and converting to float (`np.asarray(..., dtype=float)`) will read and materialise the entire raster in memory. A large or maliciously-crafted TIFF could exhaust memory/CPU and block the Dash worker. Consider enforcing maximum pixel dimensions/byte size before decoding, downsampling/tiling for display, and/or rejecting oversized uploads at the `MapLayerImport` validation layer.

_Originally posted by @Copilot in https://github.com/ImperialCollegeLondon/paricia/pull/607#discussion_r3159434945_

Sahil590:
I have added some validation to limit the file size to 100mb this can be changed in the setting file under `MAX_LAYER_FILE_SIZE_MB`. I have also added a validation for the type of file being sent. 

Fixes #608 and #618 

Would wait for #589 to be merged first


importing/models.py
```diff
@@ -222,7 +227,11 @@ class MapLayerImport(PermissionsBase):
     file = models.FileField(
         "Layer file",
         blank=False,
-        help_text="Tiff file with the layer to be imported.",
+        help_text=f"Must be a tiff file. File size must not exceed {MAX_FILE_SIZE} MB",
+        validators=[
+            FileExtensionValidator(allowed_extensions=["tif", "tiff"]),
```
dalonsoa, line 232: In addition to this, I think we should check - probably in the `clean` method - that the file can be open, contains valid geo-related data and have valid coordinates, pretty much calling [this function](https://github.com/ImperialCollegeLondon/paricia/blob/e09b782795c07e7af30ec408b79feb7029af1e13/djangomain/dash_apps/geotiff_layers.py#L100) to make sure it loads without errors, even though we are doing nothing with the outputs, in this case.

importing/utils.py
```diff
@@ -94,3 +98,8 @@ def retrieve_thingsboard_data(
         raise Exception(
             f"Failed to retrieve data: {response.status_code} - {response.text}"
         )
+
+
+def validate_layer_file_size(file) -> None:
+    if file.size > MAX_FILE_SIZE * 1024 * 1024:
+        raise ValidationError(f"File size must not exceed {MAX_FILE_SIZE} MB.")
```
dalonsoa: Add current size to the error message to inform the user in full. Also, Copilot comment seems reasonable, as well.

dalonsoa (changes requested):
I've a couple of comments, but this looks good.

importing/models.py
```diff
@@ -239,3 +239,62 @@ def __str__(self):
...
         return reverse("importing:maplayerimport_detail", kwargs={"pk": self.pk})
+
+    def clean(self) -> None:
+        """Validate the uploaded GeoTIFF and its transformed lon/lat bounds."""
+
+        if not self.file:
+            return
+
+        import rasterio
+        from rasterio.warp import transform_bounds
```
dalonsoa, line 250: Except in some circumstances, the standard is to import modules at the top of the file.

importing/models.py
```diff
@@ -239,3 +239,62 @@ def __str__(self):
...
+                        raise ValidationError(
+                            {
+                                "file": (
+                                    "GeoTIFF latitude values are out of range "
+                                    "[-90, 90]."
+                                )
+                            }
+                        )
+        except ValidationError:
+            raise
```
dalonsoa, line 293: What's the point of this? I don't feel it serves any purpose.

importing/models.py
```diff
@@ -239,3 +239,62 @@ def __str__(self):
...
+            return
+
+        import rasterio
+        from rasterio.warp import transform_bounds
+
+        file_obj = self.file.file
+        try:
+            file_obj.seek(0)
+            with rasterio.MemoryFile(file_obj.read()) as memfile:
+                with memfile.open() as dataset:
```
dalonsoa: I'm not sure I follow this logic here. `seek(0)` will go to the beginning of an open file, but you are then reading it as a `MemoryFile` and then opening it again. Could you clarify this process? reading a file before opening is very much counter intuitive.

tests/importing/test_models.py
```diff
@@ -191,11 +191,40 @@ def setUp(self):
...
+            "w",
+            driver="GTiff",
+            height=4,
+            width=4,
+            count=1,
+            dtype=np.uint8,
+            crs=CRS.from_epsg(4326),
+            transform=transform,
+        ) as dst:
+            dst.write(np.zeros((1, 4, 4), dtype=np.uint8))
```
dalonsoa, line 217: This `1` here is what indicates it is a single band, right?

tests/importing/test_models.py
```diff
@@ -220,15 +249,40 @@ def test_invalid_extension_raises(self):
...
+            buf,
+            "w",
+            driver="GTiff",
+            height=4,
+            width=4,
+            count=1,
+            dtype=np.uint8,
+            transform=transform,
+        ) as dst:
+            dst.write(np.zeros((1, 4, 4), dtype=np.uint8))
```
dalonsoa, line 284: Why don't you re-use the `_make_valid_geotiff_bytes` method you defined above?

dalonsoa (changes requested):
It's looking good, but I still have a couple of changes.

importing/models.py
Sahil590: I put it here because rasterio is quite a heavy import so wanted to avoid it being unecassarly loaded.

importing/models.py
Sahil590: Just avoids the rest being run if the file is corrupt or not detected

tests/importing/test_models.py
Sahil590: Yes it is a 1 band with a height and width of 4.

importing/models.py
Sahil590: Was left over while I was debugging this method

tests/importing/test_models.py
Sahil590: This was one of the copilots tests I left it just tests a file without a CRS and the method `_make_valid_geotiff_bytes` contains a CRS so this test would fail.

dalonsoa (approved):
LGTM!

stephwills (approved):
Sorry for not looking at this sooner -- LGTM!

dalonsoa:
Docs are failing but we have an issue for that. And link is an issue upstream that has been reported and will be sorted at some point - hopefully - so merging this one as well.

PR #625, opened 2026-05-26
Enable import of Thingsboard data

### LINKED ISSUE #613: Create a format for the thingsboard data
Currently, We can import data from things board and this automatically creates an import entry but fails due the format. 
The format is currently selected by the user. I would suggest that the format should be hardcoded when importing from Thingsboard to fit the format of the JSON submitted.

stephwills:
This PR enables Thingsboard data to be imported from its json. 

Summary
- `Format` and `Classification` are still required, but the fields related to column indices (i.e. for csv file imports) are not required. `Format` has a new `Thingsboard` field to indicate that it is for Thingsboard data. 
- The importing functions have been adapted to work for Thingsboard data. The processing is a little simpler but the user can still select the processing options for cumulative/incremental/setting a resolution.

Question
- Should it be made clearer to users which fields will be required for Thingsboard/non-Thingsboard?

Closes #613


stephwills:
@dalonsoa @Sahil590 This is just in draft for now -- for Thingsboard data, should the user still be able to specify all the processing options such as if the data is incremental, cumulative, the resolution? 

Currently, I'm just doing a plain ingestion of the data, but this misses a lot of the processing options that are available when using csv files.

dalonsoa:
To be honest, I don't know. It depends of what Thingsboard stores, if just the raw data or if it does any kind of post-processing. My feeling is that it should also be possible to define those options when importing Thingsboard data, but maybe @ICHydro has a different opinion?

stephwills:
> To be honest, I don't know. It depends of what Thingsboard stores, if just the raw data or if it does any kind of post-processing. My feeling is that it should also be possible to define those options when importing Thingsboard data, but maybe @ICHydro has a different opinion?

Hm ok I don't think Thingsboard is being used for this post-processing so it is probably good to have those options available. In which case, perhaps it is better to allow the user to define a classification (and not require all the fields related to providing column indices for csv files).

dalonsoa:
> > To be honest, I don't know. It depends of what Thingsboard stores, if just the raw data or if it does any kind of post-processing. My feeling is that it should also be possible to define those options when importing Thingsboard data, but maybe @ICHydro has a different opinion?
> 
> Hm ok I don't think Thingsboard is being used for this post-processing so it is probably good to have those options available. In which case, perhaps it is better to allow the user to define a classification (and not require all the fields related to providing column indices for csv files).

Either not required or just ignored and just used the other options of the formatting object. I agree with that suggestion.

formatting/views.py
```diff
@@ -165,6 +165,13 @@ class ClassificationCreateView(CustomCreateView):
     foreign_key_fields = ["format", "variable"]
 
 
+class ThingsboardClassificationCreateView(CustomCreateView):
+    """View to create a Thingsboard classification."""
+
+    model = Classification
+    fields = ["visibility", "variable", "accumulate", "resolution", "incremental"]
+
+
```
Sahil590: Is this supposed to used somewhere?

Sahil590 (commented):
Overall LGTM. I have tested the import process it and works fine. Some documentation about the formatting specifically about ThingsBoard would help it could be done in a different PR to avoid this one being long.

formatting/views.py
stephwills: Sorry yes, good point. At one point I started creating a separate view/form for Thingsboard specifically and abandoned it, will remove that.

dalonsoa (approved):
Aside from Sahil's comment, this looks really nice. Good work!

stephwills:
Thanks! I have added a note to the `Format` and `Classification` docstrings about the fields relevant for Thingsboard imports. Is this ok? If so, I will merge and then I can open an issue about updating the info within the documentation as that may a little more involved.

Sahil590 (approved):
LGTM!

stephwills:
I just realised I can't merge I think because of the failing docs/links. @dalonsoa are you able to merge?

PR #630, opened 2026-06-02
Mkdocs fix

Sahil590:
Was bothering me that the docs were failing so I asked copilot to see if it could see why it was failing and this is what i got seems to be working and not changing any of the documentation. 

**Explanation bellow:**

Cause: mkdocstrings-python 2.0.3 now strictly validates handler options, with members typed as `list[str] | bool | None`. Our `:::` directives used `members: None`, but YAML doesn't treat bare None as null — it parses to the literal string "None", which matches none of the allowed types and aborts the build. The correct YAML null is `null`. (Older versions didn't type-check this, so it slipped by.)

Fix: Replace `members: None` → `members: null` across all docstring directives in `docs/en` and `docs/es`. `null` parses to real `None`, which the field accepts and which means "render all members with normal filtering" — the behavior originally intended.

Verified both mkdocs build -f config/en/mkdocs.yml and config/es/mkdocs.yml build cleanly.


dalonsoa (approved):
Seems reasonable, and fixing the issue, so happy with it. Well done!

stephwills (approved):
Cool -- LGTM!

PR #641, opened 2026-06-22
Fix docs deployment

### LINKED ISSUE #356: Translate documentation in Spanish
Use mkdocs internationalisation functionality to provide the documentation in multiple languages, especially Spanish

Multi-language documentation solution.

### LINKED ISSUE #639: Fix deployment of documentation
The workflow for deploying the documentation when publishing a new release is not working, so the documentation doesn't contain the latest changes

stephwills:
This PR helps with getting the docs to deploy (I noticed the docs had not updated with the new release).

Summary:
- The current workflow to deploy the docs doesn't work because `mkdocs gh-deploy --force` doesn't work if there is no `mkdocs.yml` in the root directory.
- This PR builds the docs for each language and publishes the `site` directory to the `gh-pages` branch. 
- I've added the `index.html` that is currently being used in the `gh-pages` branch, which is copied into the `site` directory, so that the user is redirected to the EN version as default. 

Note:
- I'm not sure how to disable djlint from checking a specific fpath (rather than turning it off within the file)

Closes #639 
Closes #356

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,5 +1,6 @@
 on:
   workflow_call:
+  workflow_dispatch:
 
 jobs:
   docs:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,33 @@
+on:
+  workflow_call:
+
+jobs:
+  docs:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: 3.14
+          cache: pip
+          cache-dependency-path: requirements*.txt
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements-doc.txt
+          pip install .
+      - name: Build docs
+        run: |
+          mkdocs build -f config/en/mkdocs.yml
+          mkdocs build -f config/es/mkdocs.yml
+      - name: Copy index redirect
+        run: cp docs/index.html site/index.html
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v4
+        with:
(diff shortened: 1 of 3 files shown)


docs/index.html
```diff
@@ -0,0 +1,15 @@
+{# djlint:off #}
+<!DOCTYPE HTML>
+<html lang="en-US">
```
dalonsoa: not particularly relevant, but maybe we could use 
```suggestion
<html lang="en-GB">
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,33 @@
+on:
+  workflow_call:
```
dalonsoa, line 2: Given that you have separated this as a separate workflow - great idea! - maybe add also `workflow_dispatch`, so we can trigger the build and deployment of the documentation manually, should it be needed at some point.

dalonsoa (approved):
Many thanks for dealing with this. It was, in theory, [on me](https://github.com/ImperialCollegeLondon/paricia/issues/356), but never found time to fix it. So this PR also fixes that issue.

[According to the documentation](https://djlint.com/docs/ignoring-code/), you can disable djlint in specific lines, and for specific error codes. Have not tired, though.

Otherwise, this looks very good.

.github/workflows/docs.yml
stephwills: Good point!

stephwills:
> Many thanks for dealing with this. It was, in theory, [on me](https://github.com/ImperialCollegeLondon/paricia/issues/356), but never found time to fix it. So this PR also fixes that issue.
> 
> [According to the documentation](https://djlint.com/docs/ignoring-code/), you can disable djlint in specific lines, and for specific error codes. Have not tired, though.
> 
> Otherwise, this looks very good.

Ok thanks, I've changed it to check the specific error codes and have added `workflow_dispatch`. If it's all ok, shall I merge and then run the workflow manually?

dalonsoa:
Yes, please, go ahead

PR #644, opened 2026-07-02
Add gunicorn and caddy

### LINKED ISSUE #633: Use gunicorn to serve the website
As it is now, we are using the Django built-in `runserver` functionality, which is not really appropriate for a production environment. We should use something like `gunicorn` for that, as we do in other tools, eg. using `gunicorn paricia.wsgi:application --bind 0.0.0.0:8000` in the `docker_compose.yml` file in the server instead of `python manage.py runserver`.

The settings will be updated to properly store the media files, and the Caddy configuration will need to be changed to correctly serve them, probably to something like:

```
https://data.riverflow.io {
    # Serve media files directly
    @media {
         path /media/*
    }
    handle @media {
         root * /usr/src/app
         file_server
    }

    # Everything else goes to Django
    reverse_proxy {$HOST}:{$PORT}
}
```

stephwills:
This PR switches to using gunicorn for production purposes

Summary
- I've moved the startup command to the Dockerfile (via `entrypoint.sh`) as suggested by Adrian, which now uses `gunicorn` instead of `runserver`
  - I had to add `whitenoise` to middleware so that this would still work locally
- A Caddyfile has been added to serve the media and static files directly

I haven't configured gunicorn or caddy before so let me know if anything looks off!

Closes #633


entrypoint.sh
dalonsoa, line 1: Why do we want to dump this in a script?

docker-compose.yml
```diff
@@ -16,23 +16,38 @@ services:
       - 5432:5432
   web:
     build: .
-    command:
```
dalonsoa, line 19: Mentioned elsewhere, but what is the advantage of putting this into a script and making it the entry point of the docker image? I don't think that is how we have things setup in any other project, and I do not see any obvious advantage - or disadvantage; just don't see the reason for changing.

dalonsoa (commented):
This works OK and the changes themselves look fine, but I'm not sure about the rationale of some of the changes. If you can explain them, great, otherwise maybe @AdrianDAlessandro , who suggested the idea, can comment on that.

stephwills:
> This works OK and the changes themselves look fine, but I'm not sure about the rationale of some of the changes. If you can explain them, great, otherwise maybe Adrian D'Alessandro (@AdrianDAlessandro) , who suggested the idea, can comment on that.

Yes that is fair, I took inspiration from the direct-webapp which uses an `entrypoint.sh`, and thought it seemed cleaner to separate it out? I am happy to change it though if it doesn't make sense here.

dalonsoa (approved):
No, I think it is fine. I was just curious about why using embedding the entry point within the image. 

I guess my point for keeping the command to execute in the docker compose file is that the image is not really a standalone executable, it needs to have a database deployed and in the right port for it to work. So, executing the image with its entry point will always fail. Being that the case, why adding that capacity to the image? That is fine for DIRECT because, by default it uses a sqlite DB, so it will always run fine if you run the image. But this one will not.

stephwills:
> No, I think it is fine. I was just curious about why using embedding the entry point within the image.
> 
> I guess my point for keeping the command to execute in the docker compose file is that the image is not really a standalone executable, it needs to have a database deployed and in the right port for it to work. So, executing the image with its entry point will always fail. Being that the case, why adding that capacity to the image? That is fine for DIRECT because, by default it uses a sqlite DB, so it will always run fine if you run the image. But this one will not.

Ah ok thanks, that makes sense and I hadn't considered it when making the change! I can move it back to the docker compose or continue/merge and then continue with the migration?

I'm not sure if you had any thoughts @AdrianDAlessandro

dalonsoa:
Let's wait for @AdrianDAlessandro , but otherwise I'm happy for the merge to go ahead as it is, so you do not waste more time on this. As I said, I do not see any disadvantage, just not advantage either.

stephwills:
> Let's wait for Adrian D'Alessandro (@AdrianDAlessandro) , but otherwise I'm happy for the merge to go ahead as it is, so you do not waste more time on this. As I said, I do not see any disadvantage, just not advantage either.

Shall I merge this? I have a version of the app working with the current setup without gunicorn but want to get it configured with these changes in^

docker-compose.yml
AdrianDAlessandro: Typically, the docker-compose in our projects is for development-only and the cloud deployment doesn't include one, because it is in an Azure app service. This means the Docker image needs to include a startup command otherwise it will not do anything.

The caveat being: I don't know what the deployment setup for this app is currently. I'm unsure if it's necessary because this docker-compose is quite detailed, and also Caddy shouldn't be necessary (or even possible to use) in Azure.

djangomain/settings/settings.py
```diff
@@ -94,6 +94,7 @@
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "django_plotly_dash.middleware.BaseMiddleware",
     "djangomain.middleware.TimezoneMiddleware",
+    "whitenoise.middleware.WhiteNoiseMiddleware",
```
AdrianDAlessandro, line 97: This should be moved up to be immediately after the SecurityMiddleware. See [WhiteNoise Docs](https://whitenoise.readthedocs.io/en/stable/django.html#enable-whitenoise).

The previous production file achieved this by inserting it to index 1:
```python
MIDDLEWARE.insert(1, "whitenoise.middleware.WhiteNoiseMiddleware")
```

This is usually my preferred approach because it makes it clear what was set up by Django automatically and what is custom.

AdrianDAlessandro (commented):
Sorry for the delay in getting to this PR. There's one specific issue with where the WhiteNoise middle where is inserted that should be fixed. Other than that, it achieved what it needs to (running with gunicorn).

There was some discussion about the reasons for this approach. I've addressed one specific reason for this approach, but that's not relevant if the deployment on Azure is on a VM (which is not ideal, but simpler to set up).

And just one correction:

> I guess my point for keeping the command to execute in the docker compose file is that the image is not really a standalone executable, it needs to have a database deployed and in the right port for it to work. So, executing the image with its entry point will always fail. Being that the case, why adding that capacity to the image? That is fine for DIRECT because, by default it uses a sqlite DB, so it will always run fine if you run the image. But this one will not.

DIRECT uses a MySQL database in production, but it is not necessary for development. The configuration to link the DB to the app is done within Azure. So there are separate resources that link together (an App Service and a Database, amongst others). This approach is preferred over using a VM.

stephwills:
> Sorry for the delay in getting to this PR. There's one specific issue with where the WhiteNoise middle where is inserted that should be fixed. Other than that, it achieved what it needs to (running with gunicorn).
> 
> There was some discussion about the reasons for this approach. I've addressed one specific reason for this approach, but that's not relevant if the deployment on Azure is on a VM (which is not ideal, but simpler to set up).
> 
> And just one correction:
> 
> > I guess my point for keeping the command to execute in the docker compose file is that the image is not really a standalone executable, it needs to have a database deployed and in the right port for it to work. So, executing the image with its entry point will always fail. Being that the case, why adding that capacity to the image? That is fine for DIRECT because, by default it uses a sqlite DB, so it will always run fine if you run the image. But this one will not.
> 
> DIRECT uses a MySQL database in production, but it is not necessary for development. The configuration to link the DB to the app is done within Azure. So there are separate resources that link together (an App Service and a Database, amongst others). This approach is preferred over using a VM.

Ok thank you for this. Currently the app is on an AWS VM and is using docker compose there (with caddy). We are migrating to Azure using the App Service and a Database, and having since set that up I have realised my misunderstandings in this PR (and that Caddy isn't relevant on Azure).

I'll open a new PR to address this and move the whitenoise middleware line.

PR #646, opened 2026-07-21
Get db configuration from env vars for production

stephwills:
As we are migrating to an Azure App service + a PostgreSQL database, this PR updates the production settings so that the database connection is configured all from environment variables (currently only the password is configured this way)


dalonsoa (approved):
LGTM!

PR #648, opened 2026-07-22
Enable publish workflow to be run manually

### LINKED ISSUE #647: Modify publish workflow so deployment can incorporate latest changes
> Currently the `publish.yml` workflow, which pushes the Docker image to the GHCR, only runs when a new release is published. Should I create an issue to modify it so that it runs on `workflow_dispatch`, so it can be run manually (so the deployed version has the latest changes while I'm trying to get it working on Azure)? 

 _Originally posted by @stephwills in [#643](https://github.com/ImperialCollegeLondon/paricia/issues/643#issuecomment-5043806127)_

stephwills:
This adds `workflow_dispatch` to the publish workflow so that it can be run manually (useful for Azure deployment). 

I've also added the labels to the metadata so we can see more info, e.g. which commit it was generated from. 

I believe, according to the [metadata-action readme](https://github.com/docker/metadata-action), that the tag for a manually triggered workflow will use the branch name, i.e. `:main`, rather than `:latest`. But `:latest` and the version tags, i.e. `:v1.2.0` will still point to the published releases. Is this ok?

Closes #647

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,6 +1,7 @@
 on:
   release:
     types: [published]
+  workflow_dispatch:
 
 jobs:
   test:
@@ -33,6 +34,7 @@ jobs:
           context: .
           push: true
           tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
 
   docs:
     uses: ./.github/workflows/docs.yml
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,6 +27,12 @@ jobs:
         uses: docker/metadata-action@v6
         with:
           images: ghcr.io/${{ github.repository }}
+          flavor: |
+            latest=auto
+          tags: |
+            type=ref,event=tag
+            type=ref,event=branch
+            type=raw,value=latest,enable=${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' }}
 
       - name: Build and push Docker image for the web app
         uses: docker/build-push-action@v7


dalonsoa (approved):
This looks good to me.

About the tags, the currently deployed version uses Watchtower to check for updates to the `latest` image, so it will ignore the manually created versions as they won't have the `latest` tag, only the `branch_name` tag.

If the same approach is being used in Azure, then I would suggest to ensure that, only when triggered for the `main` branch, the tag used is `latest` and not `main`.

stephwills:
> This looks good to me.
> 
> About the tags, the currently deployed version uses Watchtower to check for updates to the `latest` image, so it will ignore the manually created versions as they won't have the `latest` tag, only the `branch_name` tag.
> 
> If the same approach is being used in Azure, then I would suggest to ensure that, only when triggered for the `main` branch, the tag used is `latest` and not `main`.

@dalonsoa Ok thanks. I haven't set it up with Watchtower currently but perhaps that is a good idea to add that.

In which case, I can make your suggested change. Would that mean that the currently deployed version will automatically update to the latest image (so I should make sure the env vars I had set for the db are added on AWS)?

dalonsoa:
The version on AWS uses Watchtower and already gets updated whenever there's a new `latest` image (unless I'm missing something - I cannot check the deployment at the moment). So it would be a question of mirroring the same configuration in Azure.

stephwills:
Ok that makes sense. Sorry yes currently it gets updated each time there is a new `latest` image, which only happens when there is a new release. 

If I make the change in this PR so that if the workflow is triggered for the main branch it uses the `latest` tag, then the version on AWS will also update to use that image? (which could introduce a breaking change if the env vars are not set up)

dalonsoa:
Update the docker compose file in AWS to use `ghcr.io/imperialcollegelondon/paricia:v1.2.0` instead of `ghcr.io/imperialcollegelondon/paricia:latest`. Essentially, you pin the version in AWS.

.github/workflows/publish.yml
```diff
@@ -26,13 +27,20 @@ jobs:
         uses: docker/metadata-action@v6
         with:
           images: ghcr.io/${{ github.repository }}
+          flavor: |
+            latest=auto
+          tags: |
+            type=ref,event=tag
+            type=ref,event=branch
+            type=raw,value=latest,enable=${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' }}
```
stephwills, line 36: My reasoning for these changes is that lines 30-34 are the default behaviour and 35 will additionally add the latest tag if a `workflow_dispatch` is made from main

stephwills:
> Update the docker compose file in AWS to use `ghcr.io/imperialcollegelondon/paricia:v1.2.0` instead of `ghcr.io/imperialcollegelondon/paricia:latest`. Essentially, you pin the version in AWS.

That makes a lot of sense thanks! I've changed the docker-compose on AWS.

dalonsoa (approved):
There's only one way of testing if this works!

Let's merge!

PR #652, opened 2026-07-31
Move whitenoise middleware

### LINKED ISSUE #651: Fix placement of Whitenoise middleware
> This should be moved up to be immediately after the SecurityMiddleware. See [WhiteNoise Docs](https://whitenoise.readthedocs.io/en/stable/django.html#enable-whitenoise).
> 
> The previous production file achieved this by inserting it to index 1:
> ```python
> MIDDLEWARE.insert(1, "whitenoise.middleware.WhiteNoiseMiddleware")
> ```
> 
> This is usually my preferred approach because it makes it clear what was set up by Django automatically and what is custom. 

 _Originally posted by @AdrianDAlessandro in [#644](https://github.com/ImperialCollegeLondon/paricia/pull/644/changes#r3667031861)_

stephwills:
This PR moves Whitenoise Middleware to directly after Django Security Middleware.

Q: I have left the caddyfile for now (even though the plan is to use Azure App service where it is not relevant) -- OK?

Closes #651

### DIFF
--- a/djangomain/settings/settings.py
+++ b/djangomain/settings/settings.py
@@ -86,6 +86,7 @@ INSTALLED_APPS = [
 
 MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
+    "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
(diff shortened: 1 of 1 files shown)


AdrianDAlessandro (approved):
All good about leaving Caddy for now. This is a more immediate thing to fix

stephwills:
> All good about leaving Caddy for now. This is a more immediate thing to fix

Ok thanks! I was thinking about whether I should address it before I tagged you for review, so that makes it easier :)

